PR #76, opened 2024-01-15
Add Phase 3 grading

prattnj:
## Backend Support for Phase 3 Grading
Changes to the backend for phase 3 include adding the passoff dependencies .jar file, adding a `PhaseThreeGrader` class not unlike the one for phases 0 and 1, and slightly modifying what is included in the classpath for the compile and run test `ProcessBuilder`s to the `server` module rather than the `shared` module. The `shared` module is still used for phases 0 and 1.

## Frontend Support for Phase 3 Grading
Only minor modifications were needed to add support for Phase 3 in the front end, such as adding another `PhaseView` and enabling the tab.

## Other
In this branch, I also updated `TestFactory.java` in all its instances to the new version, which has minimal changes. Finally, I added the default database configuration environment variables to the `Dockerfile` so they are automatically set on container startup.


Dockerfile
```diff
@@ -4,17 +4,22 @@ WORKDIR /app
...
-    yarn && \
-    yarn build
+# RUN apt-get update && \
+#    apt-get install -y curl && \
+#    curl -sL https://deb.nodesource.com/setup_21.x | bash - && \
+#    apt-get install -y nodejs && \
+#    npm install -g yarn && \
+#    cd src/main/resources/frontend && \
+#    yarn && \
+#    yarn build
```
pawlh, line 14: I think this should stay enabled by default so that Docker remains an option for production deployments

src/main/java/edu/byu/cs/autograder/PassoffTestGrader.java
```diff
@@ -45,6 +50,11 @@ public PassoffTestGrader(String phaseResources, String netId, String repoUrl, Ob
         super(repoUrl, netId, observer, phase);
         this.stageTestsPath = new File(stagePath + "/tests");
         this.phaseTests = new File(phaseResources);
+        this.module = switch (phase) {
+            case Phase0, Phase1 -> "shared";
+            case Phase3, Phase4 -> "server";
+            case Phase6 -> "client";
```
pawlh, line 56: I think Phase 6 grading will need the server jar, but we might also need the client for checking unit tests. We can cross this bridge later though

src/main/java/edu/byu/cs/autograder/PassoffTestGrader.java
```diff
@@ -107,19 +118,20 @@ protected TestAnalyzer.TestNode runTests() {
...
                         "-jar",
                         standaloneJunitJarPath,
-                        "--class-path", ".:" + chessJarWithDeps + ":" + junitJupiterApiJarPath,
+                        "--class-path",
+                        ".:" + chessJarWithDeps + ":" + junitJupiterApiJarPath + ":" + passoffDependenciesPath,
                         "--scan-class-path",
                         "--details=testfeed");
 
         try {
+            System.out.println("Compiled command: " + String.join(" ", processBuilder.command().toArray(new String[0])));
```
pawlh, line 134: For noise purpose I think this should be a `LOGGER.info`

pawlh (changes requested):
The only big thing that might hold us back here are the unit tests they have to write. If we can make it so that this only updates a part of the rubric, we could put that part off until later

Dockerfile
prattnj: I forgot to uncomment that, that's on me.

src/main/java/edu/byu/cs/autograder/PassoffTestGrader.java
prattnj: Another thing I meant to remove, oops. Thanks for noticing!

PR #123, opened 2024-01-25
118 backend deny submissions when the student has earned max score

DallinFromEarth:
prevents the grader from running if the student already has earned the max score

changes include:
- refactoring some code in SubmissionController so that the most recent submission isn't run across duplicate code
- running a check before the grader runs, querying the most recent submission for the phase

### DIFF
--- a/src/main/java/edu/byu/cs/controller/SubmissionController.java
+++ b/src/main/java/edu/byu/cs/controller/SubmissionController.java
@@ -123,11 +123,11 @@ public class SubmissionController {
     };
 
     /**
-     * checks to see if the specified student has already achieved the highest possible grade on the specified phase
+     * checks to see if the specified student achieved the highest possible grade on the specified phase on their most recent submission
      * @param netId netId of the student to check
      * @param phase phase of the project to check
-     * @return true if the student already has the highest possible grade.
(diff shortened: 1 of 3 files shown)


src/main/java/edu/byu/cs/controller/SubmissionController.java
```diff
@@ -82,12 +82,12 @@ public class SubmissionController {
...
         }
-        SubmissionDao submissionDao = DaoService.getSubmissionDao();
-        Submission submission = submissionDao.getSubmissionsForPhase(user.netId(), request.getPhase()).stream()
-                .filter(s -> s.headHash().equals(headHash))
-                .findFirst()
-                .orElse(null);
-        if (submission != null) {
+        if (mostRecentHasMaxScore(netId, request.getPhase())) {
+            halt(400,"You have already earned the highest possible score on this phase");
+            throw new RuntimeException("Max score already earned on this phase");
```
pawlh, line 87: Sorry, I know I initially said to throw a runtime exception, but I think this would should instead be `return null;` to match the other guards in this class

DallinFromEarth: consider it done, because its done

pawlh (approved):

PR #138, opened 2024-01-31
Remove syncronized

pawlh:


frozenfrank:
The `synchronized` keyword was causing deadlock when a null-pointer exception was thrown from the target function.

PR #139, opened 2024-01-31
add timeout for processes

pawlh:


frozenfrank (commented):
Yup! We worked on this together, and it all looks good.

Issue #143, opened 2024-01-31
frontend: Sort submissions by submitted date DESC


frozenfrank:
Or, more generally, we could swap out the table for a library that would allow us to dynamically sort the table by any of the columns. [Ag-Grid](https://www.ag-grid.com/) is a great option.

Issue #112, opened 2024-01-23
frontend: fix frontend AM/PM display


pawlh:
A submission made at 12:24 PM is incorrectly showing as 12:24 AM

![Image](https://github.com/softwareconstruction240/autograder/assets/10139415/05378a4d-2df3-4aa7-85d3-1d112712065a)

frozenfrank:
This is currently being addressed by using a native date formatting method. If we want a more flexible option that allows us to format the dates in more custom ways, I recommend [MomentJS](https://momentjs.com/).

frozenfrank:
Note that this bug is currently also present in all timestamps across the app, including in the submissions list. Here is a screenshot of the issue in another place:

<img width="750" alt="image" src="https://github.com/softwareconstruction240/autograder/assets/19808440/fc8b763c-c137-463b-83dd-e279a8ef8527">

PR #146, opened 2024-02-01
front-end: multiple improvements

frozenfrank:
This addresses multiple issues:
* #143 
* #112 
* Submission table efficiency issue
* Easy of using `readableTimestamp(Date|string)` method

Note that while multiple changes are squashed into this single PR, the commit history has been intentionally authored. Each of the changes is contained in a separate commit for individual consideration and inspection.

The reviewer of my code will definitely want to run and review my code before accepting this PR. I don't have my machine setup to test and run the code for actual correctness, though the ideas contained in it are good.

Specifically, it would be good to verify that the timestamp display issue is properly corrected by calling the native method.


src/main/resources/frontend/src/stores/admin.ts
```diff
@@ -5,9 +5,18 @@ import {usersGet} from "@/services/adminService";
...
+    const usersByNetId = ref<Record<string, User>>({});
 
     const updateUsers = async () => {
-        users.value = await usersGet();
+        const latestUsers = await usersGet();
+        users.value = latestUsers;
+
+        // Also store a cache of all the users by their netId
+        const byNetId: typeof usersByNetId = {};
+        for (let user of latestUsers) {
```
pawlh, line 16: TS demands some gymnastics here. `typeof usersByNetId` is actually `Ref<Record<string, User>>: value, [RefSymbol]` instead of just `Record<string, User>`

I'm not sure how to dig the inner type out of Refs in Vue, so using a common interface (or just duplicated type defs) would probably be a good solution

src/main/resources/frontend/src/stores/admin.ts
pawlh, line 1: `usersByNetId` needs to be exposed by the store, so the return line should be changed to `return {users, usersByNetId, admins, students, updateUsers}`

pawlh (changes requested):

src/main/resources/frontend/src/stores/admin.ts
frozenfrank: Ah! Good point. I'm not familiar with `vite` to recognize the difference between their syntax and standard `TypeScript`. Thanks for the tips— I'll adjust it

pawlh (approved):

pawlh:
Looks great! Thanks for doing this!!

PR #147, opened 2024-02-01
Add Caching to Commit Analytics

prattnj:
On the Analytics tab, admins now have the option to download a cached version of the commit data rather than having to wait 5 minutes for the new data. They still have the option to update the data if desired. If there is no cached data (this should only be the case until it is calculated the first time), the button is disabled. Additionally, a `util` package was added and some files underwent minor refactoring.


src/main/java/edu/byu/cs/util/DataUtils.java
```diff
@@ -0,0 +1,28 @@
...
+    public static Submission getFirstPassingSubmission(String netID, Phase phase) {
+        Collection<Submission> submissions = DaoService.getSubmissionDao().getSubmissionsForPhase(netID, phase);
+        Submission firstPassing = null;
+        int earliest = Integer.MAX_VALUE;
+        for (Submission s : submissions) {
+            if (!s.passed()) continue;
+            if (s.timestamp().getEpochSecond() < earliest) firstPassing = s;
+        }
+        return firstPassing;
+    }
```
pawlh, line 27: I think this should be a Dao method instead. I know you didn't write this initially, but I don't think it belongs in a class called `DateUtils`. If you'd rather not change it in this PR, could you open a new issue for it? #cleancode 😃

src/main/java/edu/byu/cs/util/PhaseUtils.java
```diff
@@ -0,0 +1,93 @@
...
+        };
+    }
+
+    /**
+     * Given an integer representation of a phase as a string, returns the phase
+     *
+     * @param str examples include "0", "3"
+     * @return the phase as an enum
+     */
+    public static Phase getStringAsPhase(String str) {
```
pawlh, line 52: How about `getPhaseByString` instead? That matches existing repo conventions

src/main/resources/frontend/src/services/adminService.ts
```diff
@@ -100,9 +100,9 @@ export const getQueueStatus = async (): Promise<QueueStatusResponse> => {
     }
 }
 
-export const commitAnalyticsGet = async (): Promise<string> => {
+export const commitAnalyticsGet = async (option: string): Promise<string> => {
```
pawlh, line 103: Could you add a Typescript definition containing the possible values for option?

src/main/resources/frontend/src/views/AdminView/Analytics.vue
```diff
@@ -1,23 +1,56 @@
...
+  await getData('Downloading cached data...', 'cached', false)
+}
+
+const getNewData = async () => {
+  await getData('Downloading data... (should take around 5 minutes) Enjoy this 5-minute song in the meantime.',
+      'update', true)
+  await getMostRecent()
+}
+
+const getData = async (info: string, option: string, music: boolean) => {
```
pawlh, line 39: This option would also share the type from `adminService`

src/main/resources/frontend/src/views/AdminView/Analytics.vue
```diff
@@ -1,23 +1,56 @@
...
+  updateButtonDisabled.value = true
+  cachedButtonDisabled.value = true
+  infoText.value = info
+  if (music) playAudio()
```
pawlh, line 43: ❤️

src/main/java/edu/byu/cs/analytics/CommitAnalyticsRouter.java
```diff
@@ -0,0 +1,57 @@
...
+     *
+     * @return the CSV data
+     */
+    public static String update() {
+        long ts = Instant.now().getEpochSecond();
+        String data = CommitAnalytics.generateCSV();
+
+        FileUtils.removeDirectory(new File(cacheDir));
+        FileUtils.createDirectory(cacheDir);
+        FileUtils.writeStringToFile(data, new File(cacheDir + "/" + ts + ".csv"));
```
pawlh, line 27: Just checking, will anything catastrophic happen if the file can't be created? It might be a good idea to create a specific error logging line for this

pawlh (changes requested):

src/main/java/edu/byu/cs/analytics/CommitAnalyticsRouter.java
prattnj: If the file can't be created, a RuntimeException is thrown and the client gets the message, so nothing catastrophic. I still added a line to log the error, however, that's a good idea.

src/main/java/edu/byu/cs/util/DataUtils.java
prattnj: Done

pawlh (approved):

pawlh (approved):
Sorry for the delay... looks good!

PR #153, opened 2024-02-06
121 backend add canvas rubric submissions

19mdavenport:


src/main/java/edu/byu/cs/autograder/PhaseThreeGrader.java
```diff
@@ -16,4 +16,11 @@ public class PhaseThreeGrader extends PassoffTestGrader {
     public PhaseThreeGrader(String netId, String repoUrl, Observer observer) throws IOException {
         super("./phases/phase3", netId, repoUrl, observer, Phase.Phase3);
     }
+
+    /* Rubric Items Winter 2024:
```
frozenfrank, line 20: Are you sure you meant to include these time-specific comments in the code? It would seem prudent to remove them if possible, or at least include the full details if it's necessary that they continue. (It looks like the beginning portion of some of the fields have been omitted.)

A better solution would be to include some documentation in the `README.md` file that shows where the actual rubrics should be managed. Since this is commented out, I'm assuming that the actual values are stored in a table somewhere. That should be described.

src/main/java/edu/byu/cs/canvas/CanvasIntegration.java
```diff
@@ -151,19 +153,78 @@ public static void submitGrade(int userId, int assignmentNum, float grade, Strin
...
+    public static void submitGrade(int userId, int assignmentNum, Map<String, Float> grades, Map<String, String> rubricComments, String assignmentComment) throws CanvasException {
+        CanvasSubmission submission = getSubmission(userId, assignmentNum);
+        if(submission.rubric_assessment() != null) {
+            for(Map.Entry<String, CanvasIntegration.RubricItem> entry : submission.rubric_assessment().items().entrySet()) {
+                grades.putIfAbsent(entry.getKey(), entry.getValue().points());
+                rubricComments.putIfAbsent(entry.getKey(), entry.getValue().comments());
+            }
+        }
+
+        StringBuilder queryStringBuilder = new StringBuilder();
```
frozenfrank, line 176: It seems like this block of code from line `176` to `191` would do well to be extracted into a separate function. Based on my interpretation of the code as it stands, it seems that a reasonable name could be `buildSubmissionQueryURL()`. It would probably be best to have the function handle **all** of the string building, whereas right now, some is handled in these control blocks, and some occurs via direct string concatenation on line `195`. That would do a lot to improve the readability of the code.

src/main/java/edu/byu/cs/canvas/CanvasIntegration.java
```diff
@@ -151,19 +153,78 @@ public static void submitGrade(int userId, int assignmentNum, float grade, Strin
...
+     * @param rubricComments    A Map of rubric item id's to comments to put on that rubric item
+     * @param assignmentComment A comment for the entire assignment, if necessary
      * @throws CanvasException If there is an error with Canvas
      */
-    public static String getGitRepo(int userId) throws CanvasException {
-        CanvasSubmission submission = makeCanvasRequest(
+    public static void submitGrade(int userId, int assignmentNum, Map<String, Float> grades, Map<String, String> rubricComments, String assignmentComment) throws CanvasException {
+        CanvasSubmission submission = getSubmission(userId, assignmentNum);
+        if(submission.rubric_assessment() != null) {
+            for(Map.Entry<String, CanvasIntegration.RubricItem> entry : submission.rubric_assessment().items().entrySet()) {
```
frozenfrank, line 170: This code is smelly to me.

This control block is mutating the `grades` and `rubricComments` data structures that were originally passed in from the user based on data extracted from other portion of the app. This is scary because the callers of this function may not understand or expect that behavior (a short time after the original authorship of the code). 

If it's strictly necessary to merge in comments from these two data sources, then that should be explicitly stated and explained. However, it's likely that there is another way of accepting the comments that does one of the following:
1. Requires the user to acknowledge & control the merging process by explicitly calling a function or setting a flag
2. Collects the data from various stages of calculation without a required merging process (why are there multiple places to look anyways)?

src/main/java/edu/byu/cs/canvas/CanvasIntegration.java
```diff
@@ -138,6 +139,7 @@ private static Collection<User> getMultipleStudents(String baseUrl) throws Canva
      * @param userId        The canvas user id of the user to submit the grade for
      * @param assignmentNum The assignment number to submit the grade for
      * @param grade         The grade to submit (this is the total points earned, not a percentage)
+     * @param comment       The comment to submit on the assignment
```
frozenfrank, line 142: Celebration 🎉

Improved JavaDoc documentation is incredible for everyone!

frozenfrank (changes requested):
Assuming that this code works, I would want it code-smithed into a more readable format before actually merging it the codebase. I left specific comments on areas that would benefit from this.

As I understand it, the goal here is to build an AutoGrader that will outlast us and continue it's service into future semesters. The best way to prop it up for success is by writing clean, understandable code from the beginning that future people will be able to work without unknowingly causing bugs.

src/main/java/edu/byu/cs/autograder/PhaseThreeGrader.java
19mdavenport: There is no table that you assumed exists

frozenfrank: @19mdavenport Well then I'm even more confused. What is the purpose of including these rubric items in this comment?

19mdavenport: It was just a help for whoever is doing the phase 3 grader code so they don't have to look it up again

prattnj (approved):

prattnj (approved):

frozenfrank (approved):
Thanks for clarifying why the code is behaving the way it is. This looks good!

PR #155, opened 2024-02-08
126 use full date times for dates

### LINKED ISSUE #126: backend: use full datetime for due dates
Currently the due date checker ignored the hour, minutes, and seconds. The canvas time should be canonical

frozenfrank:
This PR resolves #126 and also improves the quality of the code in the area directly related to that issue.

Please run and test my code before merging since I don't have a way to run the code locally.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/Grader.java
+++ b/src/main/java/edu/byu/cs/autograder/Grader.java
@@ -162,19 +162,7 @@ public abstract class Grader implements Runnable {
 
         int canvasUserId = DaoService.getUserDao().getUser(netId).canvasUserId();
 
-        ZonedDateTime dueDate;
-        try {
-            dueDate = CanvasIntegration.getAssignmentDueDateForStudent(canvasUserId, assignmentNum);
-        } catch (CanvasException e) {
-            throw new RuntimeException("Failed to get due date for assignment " + assignmentNum + " for user " + netId, e);
-        }
-
-        // penalize at most 5 days
-        ZonedDateTime handInDate = DaoService.getQueueDao().get(netId).timeAdded().atZone(ZoneId.of("America/Denver"));
-        int numDaysLate = Math.min(DateTimeUtils.getNumDaysLate(handInDate, dueDate), 5);
-        float score = getScore(results);
-        score -= numDaysLate * 0.1F;
-        if (score < 0) score = 0;
+        float score = getFinalAdjustedScore();
 
         SubmissionDao submissionDao = DaoService.getSubmissionDao();
         Submission submission = new Submission(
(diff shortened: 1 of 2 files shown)


src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -197,6 +185,39 @@ private void saveResults(TestAnalyzer.TestNode results, int numCommits) {
...
     }
 
+    /**
+     * Gets the reported score from the grader, and then adds appropriate adjustments.
+     * These adjustments include any late penalty that should be applied.
+     *
+     * @return The final score that should be saved in the Gradebook.
+     */
+    private float getFinalAdjustedScore() {
+        final int MAX_PENALIZE_DAYS_LATE = 5; // Lose credit for days late UP TO this value
```
frozenfrank, line 195: For the reviewer:

This code is already more readable than the existing code, but some question may be asked about why these constants are included inside this function. Another option would be to include them in a more global place where they could be easily changed or adapted in the future. I decided **not** to do this because I don't know in which global place I would put them. Additionally, I'm not certain these are the variables that will make the difference (future changes to this code may require changing these variables anyways). Therefore, I put them close to the code they adjust.

I'm happy to move them to a different location, or you can do it yourself since I gave reviewers ability to adjust the PR.

pawlh: Agreed. In several places I've added `// FIXME: ...` lines above magic numbers to remind us to go back through later and pull them out to be in a more reasonable spot, but it looks like I forgot to note this one. Thanks for pointing it out!

There are a number of configurations that should probably live in some sort of `course_settings` table. Off the top of my head this would include things like the Canvas service account API key, holidays, and this

src/main/java/edu/byu/cs/util/DateTimeUtils.java
```diff
@@ -31,9 +31,6 @@ public static String getDateString(long timestamp, boolean includeTime) {
      * @return the number of days late or 0 if the submission is not late
      */
     public static int getNumDaysLate(ZonedDateTime handInDate, ZonedDateTime dueDate) {
-        // end of day
-        dueDate = dueDate.withHour(23).withMinute(59).withSecond(59);
-
```
pawlh, line 36: Thanks for remembering this from the TA meeting; I forgot to follow through with it. We might have a problem with at the current moment though, as it looks like the due dates on Canvas are still set to 8pm
image

pawlh: I'm definitely not opposed to letting Canvas be the source of truth, but we have quite publicly promised that students have until midnight, so I don't want to include this change until Canvas is correct

pawlh: I'm definitely not opposed to letting Canvas be the source of truth, but we have publicly promised that students have until midnight so I don't want to include this change until Canvas is correct

pawlh (changes requested):
Let's either put back the 11:59:59 pm rule, or wait to merge this until we have confirmed that Canvas is how the professors want it

frozenfrank:
@pawlh Since you aren't actually "requesting changes," can you comment or approve the PR? There is a distinction between having the PR "approved" and actually merging it in to main. 

I'm on board with waiting until Canvas is updated before we merge the code as evidenced by [this Slack Message](https://winter2024cs240tas.slack.com/archives/C06CX61QUFK/p1707429233581369).

src/main/java/edu/byu/cs/autograder/Grader.java
frozenfrank: @pawlh See #156 which tracks the creation of a `course_settings` table.

pawlh:
> @pawlh Since you aren't actually "requesting changes," can you comment or approve the PR? There is a distinction between having the PR "approved" and actually merging it in to main.
> 
> I'm on board with waiting until Canvas is updated before we merge the code as evidenced by [this Slack Message](https://winter2024cs240tas.slack.com/archives/C06CX61QUFK/p1707429233581369).

I'd argue that this was an appropriate use of requesting changes. In this case there were two options for changes, and action needed to be taken on one of them before I would be comfortable approving anything for merging:
1. Remove the problematic lines of code
2. Change the environment (Canvas)

Until a change was made the code would cause an undesired result. In a CI/CD situation (which I desperately wish were the case here) the `main` branch and the production environment are closely connected. Still, I'll admit that this was probably an uncommon usage of the "Request Changes" feature

Either way, action was taken and my concern is mitigated, so I'll approve this now. Thanks for doing this!

pawlh (approved):

Issue #166, opened 2024-02-10
backend: Clean up requests for DB credentials


frozenfrank:
Take ear to the following observations:
* In the setup instructions, there is a reference to a `src/main/resources/db.properties` file 
* That file does not exist in the current repository
* The "Environment Variables" portion of root README.md file shows environment variables that define the DB username, password, url, and name
* There are two other files that contains this information:
    * `phases/phase4/resources/db.properties`
    * `phases/phase6/resources/db.properties`
* Those other files seem to be stale references to old settings

Based on the information currently available, I propose that we remove these outdated files and rely wholly on the environment variables. The README.md file should be updated to reflect these changes.

Note that there may be legitimate concern for database access with later phases. I'm not informed enough about how those files are supposed to access student's databases. Please comment with any relevant information.

pawlh:
Yes to half of this.

The reference in `README.md` to `src/main/resources/db.properties` is outdated and should be removed.

As for the `db.properties` files in the `phases/` path, those are different and mostly not related to the autograder itself. We will need those in order to run students' code. We might not end up just copying them in, but the specific procedure has not yet been worked out. (see the instructions for Phase 4, step 1: https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/4-database/database.md)

frozenfrank:
@pawlh Thanks for the information! I'll save the notes about the `db.properties` for later, but I have addressed this concern:

> The reference in README.md to src/main/resources/db.properties is outdated and should be removed.
This is handled in #167.

pawlh:
I've lost track of which PR it was, but I believe this has been addressed by @19mdavenport's db changes that dynamically generate student db info for each grading session

Issue #165, opened 2024-02-10
backend: Provide mechanism for PROD vs DEV compilations


frozenfrank:
In other software systems, I've had access to an `environment.production` flag that we use to enable some behaviors for development only. The key distinction is that this flag is automatically set to `true` when the app is built for production deployment but set to false everywhere else which removes the manual hassle of changing (or forgetting) to change the flag. 

Some deployment platforms already have built-in mechanisms for this kind of behavior. In the Autograder, this could be achieved by adding a `production` flag to the `config.properties.java` file; a special build step would need to set it to `true` when building for PROD though.

This issue relates to the following other features:
* #163 
* #164

pawlh:
This would be great.

The frontend is already mostly set up for this if a `.env.development` file is present in the root and it is run with `yarn dev`, but the backend currently has no notion of this. I've never explicitly done this with Java before, so I'm curious what the general conventions are. I'm guessing that your suggestion is the right way to go

frozenfrank:
No be fair, I’ve only ever done it with front end libraries as well.
Angular also provides a `environment.ts` file and an
`envirnoment.production.ts` file that it swaps out appropriately.

As for doing it in the back end, I’ll have to research it some more. An
optional environment variable could represent PROD, or we could use a file
swap like frameworks do for us automatically…

Let’s keep the conversation going on this topic.

Issue #164, opened 2024-02-10
backend: Provide mechanism of locally validating as ADMIN without Canvas


frozenfrank:
Currently, the Autograder requires a Canvas Authorization Key to look up the signed in user to determine if they are admins or not. While this works great for production, for development, it requires setting up keys that may not need to be necessary.

It would be great if the autograder would automatically allow me to be an ADMIN when it's not running in production mode. While implementing this request, we should be careful not to disallow the current flows to still view the app as a test student. The Autograder should _allow_ me to be an ADMIN, and it should also allow me to be a _STUDENT_ if I want.

pawlh:
I like this idea, but it would definitely fundamentally change how authentication work.

Originally (and before we released the autograder to students) people were authenticated with CAS then redirected to a page where they could manually register. We changed this in favor of having a persons's Canvad profile tightly coupled with their autograder profile. Changing this is certainly doable, but I'm a little skeptical of its utility.

> it requires setting up keys that may not need to be necessary

I'm open to changing my mind here. What are some specific reasons that requiring an auth key during development is sub-optimal?

frozenfrank:
I think this is the issue where we talked about with the Canvas Integration classes being converted into an Interface. @19mdavenport Has begun the work on this in #260, but we'll need to continue that work by explicitly verifying that the mock implementation of the interface works as expected in all cases.

Issue #163, opened 2024-02-10
backend: Create memory DAO's that allow us to skip MySQL for development


frozenfrank:
When just trying to test front-end stuff, and I don't really need the SQL server. I should be able to run the server in `dev` mode that will use memory DAO's instead of the database DAOS. Ideally, it would also automatically populate the DAO's with some data that is helpful for having an environment.

This would be well accomplished by providing alternatives in the `DaoService` file that ends up using memory DAO's instead of Database DAO's.

pawlh:
I absolutely agree with this one. You might have noticed that I have half-hearted support for a memory database already set up, but there is not currently an easy way to enable it.

This is something that has been on my todo, so thanks for bringing it closer to reality with this issue :)

ThanGerlek:
Do we want to try to make this work with phases 4-6? I've created a dev branch with proposed implementations of the other two DAOs (including pulling rubric configs from a json file) (no actual injection yet though), which is enough for phases 0-3 as long as we only submit repos that are themselves set to use memory DAOs. Obviously not ideal but probably enough for any front-end testing.

PR #167, opened 2024-02-10
Add and update getting started instruction

frozenfrank:
These instructions reflect the process that @19mdavenport lead me through in order to get started with the project on my local machine. As of yet, I haven't attempted to _grade_ any projects, but only to run the front-end and back-end for other reasons. With that said, the process may not be complete if my code needs to grade code.

While I updated everything that I knew, there were other issues that I didn't know for myself. I've created separate issues (namely #166) to continue the improvement of documentation.

Additionally, while getting set up, there were other unnecessary hurdles that could be minimized in the future. Those have also been noted as separate issues (namely #163, #164, and #165).

Other than that, I feel I did a good job capturing the information that a new person would need to follow to build the project on their own machine in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,31 +41,70 @@ Note, if you are using the docker MySQL database, ensure that in `src/main/resou
 ## Development
 ### Pre-requisites
 #### Node/Yarn
-The frontend is built using Vue.js. To run the frontend, you will need to have `yarn`. After installing `Node`, run the following to enable `yarn` globally:
+The frontend is built using Vue.js. To run the frontend, you will need to have `yarn`.
+After installing `Node`, run the following to enable `yarn` globally:
 ```bash
 corepack enable
 ```
 (`sudo` may be required)
 
-The frontend can be run in development mode by running the following in the `src/main/resources/frontend` directory:
-```bash
-yarn dev
-```
+#### Backend
+A local installation of maven is required to run the backend.
 
-#### Docker
-To run the database locally, you will need to have `docker` installed. Run the following in the root of the project to start the db container:
+Alternatively, you can run the database locally with your own MySQL server.
+Be sure to update `src/main/resources/db.properties` with the correct database url, username, and password.
+
+#### Database
+To run the database locally, you will need to have `docker` installed.
+
+Alternatively, you can run the database locally with your own MySQL server.
+Be sure to update `src/main/resources/db.properties` with the correct database url, username, and password.
+
+##### Docker
+Run the following in the root of the project to start the db container:
 ```bash
 docker compose up db -d
 ```
 
-#### Maven
-A local installation of maven is required to run the backend.
+##### MySQL
+Go fish🐟
 
+_This is required, but no setup instructions are given in this document._
 
-Alternatively, you can run the database locally with your own MySQL server. Be sure to update `src/main/resources/db.properties` with the correct database url, username, and password.
+### Getting Started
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,31 +41,70 @@ Note, if you are using the docker MySQL database, ensure that in `src/main/resou
...
-#### Maven
-A local installation of maven is required to run the backend.
+##### MySQL
+Go fish🐟
 
+_This is required, but no setup instructions are given in this document._
 
-Alternatively, you can run the database locally with your own MySQL server. Be sure to update `src/main/resources/db.properties` with the correct database url, username, and password.
+### Getting Started
+NOTE: These instructions will help you set up this project in Intelli-J.
```
frozenfrank, line 75: How do we feel about specializing on **Intelli-J**? Do we want to be more general and support VSCode as well? Or, like the attitude we're taking in the class, are we OK with helping people adopt the system of _this_ organization which is Intelli-J? (That's the approach I assume we're taking, and I assume we're ok with it. Furthermore, it's the only information I have at the this point.)

frozenfrank:
Also note that I adjusted the headline depths of some of the existing headers. These have been intentionally crafted to read as a document better, and to keep information closest to the topic to which it is related.

This results in the following overall file structure which plays well with code folding features:
```markdown
# Autograder
## Important Folders
...
## Deployment
...
## Development
...
```

README.md
```diff
@@ -41,31 +41,66 @@ Note, if you are using the docker MySQL database, ensure that in `src/main/resou
...
+##### MySQL
+Go fish🐟 These instructions are not included in this file.
 
+### Getting Started
+NOTE: These instructions will help you set up this project in Intelli-J.
+If there are any holes or gaps in the instructions, please submit a new pull request
+to share the learned knowledge with the people behind you.
+1. Clone the repo onto your local machine.
+2. Ensure that you are using the latest version of JVM. The system currently requires version 21+.
+3. Use `yarn` to install all dependencies. This only requires running `yarn` in the root directory.
```
pawlh, line 76: It'd probably be worth noting that in this case, `root` is defined as `src/main/resources/frontend`

README.md
```diff
@@ -41,31 +41,66 @@ Note, if you are using the docker MySQL database, ensure that in `src/main/resou
...
+      - Put them all onto a single line, with each variable separated by a `;` (semicolon)
+   - Edit the "Server" run configuration
+      - Paste in the single line of parameters to the line titled "Environment Variables"
+      - Going forward, there is an option to open up a table of the variables where you can
+        edit just one value from the lot
+   - Save & apply the changes
+6. **Enable logging** by following the instructions below
+7. **Change config properties**
+   - Navigate to the following file: `src/main/resources/config.properties`
+   - For a simple development setup, skip now to the subheading **Dev Config Properties**
```
pawlh, line 93: I couldn't find a subheading with this title, nor any section that gives suggested values for the properties in `config.properties`

pawlh (changes requested):
Just a couple things. Otherwise it looks great, thanks!!

README.md
pawlh: I think it's fine to expect IntelliJ. I do not think it would be very difficult for somebody to adopt this to work for VSCode or some other platform

frozenfrank:
I pushed new commit that address the two concerns you described in the comments.

pawlh (approved):

Issue #156, opened 2024-02-08
backend: Store Configurable values inside a course_settings table


frozenfrank:
## Description
Some kinds of parameters we want to be configure dynamically. This can we well accomplished with a `course_settings` table that has a column for each of the settings we want to configure. Then, the app will read out the appropriate settings and behave differently just by updating the values _but without rebuilding the app_.

In the code, we'll want to access these through a `CourseSettingsService` (or similarly named) by the name of the property. This would allow us to potentially configure different sets of values that we could switch between. For example, one row of settings could contain the PRODUCTION settings, and another row could contain DEV settings which have have slightly different end points or behaviors for some properties. We could also build in a file of defaults so that we aren't required to define all of them in the database if we want.

 ## Properties
* Canvas API Key
    * Canvas Test Student ID
* Assignment ID's for the different phases
    * Rubric item IDs for various rubric items
* Late penalty configuration
    * Late penalty percent per day
    * Late penalty max days
    * List of Holidays
* Course Slack API details for sending messages (future) 
* etc...

Please comment any other properties that would be good to include in the `course_settings` table.

19mdavenport:
The Canvas Test Student ID resets every time someone hits "reset student." I think the way Paul added that gets it new every time we need it is probably sufficient.

Canvas API Key is currently passed in as a environmental variable, do we need one per course?

pawlh:
> Canvas API Key is currently passed in as a environmental variable, do we need one per course?

I don;t know about per course, but if #164 happens then I think it would be super useful to have this in the database instead

pawlh:
Another class of properties that should be included:

- weights of rubric items. e.g. _**PHASE_3_UNIT_TESTS=.3**_, _**PHASE_3_PASSOFF_TESTS=.7**_, etc.

pawlh:
I added this with #323, but I'm not going to close this ticket because it can probably improved on

PR #173, opened 2024-02-17
168 expected penalties are applied to grade after extra credit points are added

### LINKED ISSUE #168: Expected penalties are applied to grade after extra credit points are added
## Overview
Adjust the calculation of final scores so that any late penalties (or other penalties applied for low commit frequency etc) are applied based on a percentage of the final score _after_ any extra credit points are accounted for.

The grade should be calculated as the following formula:
```
LATE_PERCENTAGE_PER_DAY = 10
LOW_COMMIT_PENALTY_PCT = 10

gradedScore = percentage_tests_passed + extra_extra_credit_points_available
latePenaltyModifier = 1 - (LATE_PERCENTAGE_PER_DAY/100 * effective_days_late)
lowCommitModifier = 1 - (commits_under_permissible_threshold ? 0 : LOW_COMMIT_PENALTY_PCT)
finalScore = gradedScore * latePenaltyModifier * lowCommitModifier

return finalScore;
```

## Discussion
Professor Jensen said the following:
> I would have expected penalties (late, git commits) to apply after calculating the final score.

This was said in response to the following TA discussion
> **TA 1:** Should the late penalty apply to the extra credit points? It looks like the autograder is not currently doing that. For example, a student got 78% turning it in 3 days late, rather than 75.6% I would expect if the late penalty applied to those points

> **TA 2:** Personally I don't think it should penalize the extra credit, although I'd agree that the syllabus wording isn't perfectly clear: "The late penalty for all assignments is 10% off per school day"

> **TA 3:** This [proposed adjustment] would also be consistent with the behavior for partial credit submissions which may occur on phase 6.

##  Test Cases
* A student passes all the tests and the extra credit tests for a total raw score of **108%**. However, the student turns in the code **one school day** late (10% per day). This results in a final score of **97.2%** 
* On a later phase, a student receives credit for **60%** of the points on the rubric. However, the student only committed on **2 days** instead of the 3 separate days required by #160. Therefore, a 10% penalty is applied for a total score of **54%**.
* A student passes all of the test cases, but turns in the code **8 school days** late. He is only penalized for the first 5 of these, so the final score if **50%**.

frozenfrank:
This closes  #168.

It also begins to put some of the key functionality in place for #156 via a `GradingSettings` record.

While I have confirmed that this code compiles, I'm not sure how to run it on submissions to verify that the results are correct. Please do that, or help me figure out how to.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/Grader.java
+++ b/src/main/java/edu/byu/cs/autograder/Grader.java
@@ -239,7 +239,7 @@ public abstract class Grader implements Runnable {
     }
     private int getDaysLate(ZonedDateTime dueDate, ZonedDateTime handInDate) {
         int actualDaysLate = DateTimeUtils.getNumDaysLate(handInDate, dueDate);
-        // FIXME! Exclude certain holidays
+        // FIXME! Exclude certain holidays and weekends
         return Math.min(gradingSettings.MAX_PENALIZE_DAYS_LATE(), actualDaysLate); // Effective days late
     }
     private float applyGradePenalties(float reportedScore, int daysLate) {
--- a/src/main/java/edu/byu/cs/autograder/Grader.java
+++ b/src/main/java/edu/byu/cs/autograder/Grader.java
@@ -87,6 +87,10 @@ public abstract class Grader implements Runnable {
      * The required number of commits (since the last phase) to be able to pass off
      */
     private final int requiredCommits;
+    /**
+     * Holds configurable settings related to the grading of assignments.
+     */
+    private final GradingSettings gradingSettings;
 
     protected Observer observer;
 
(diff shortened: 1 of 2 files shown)


frozenfrank:
@19mdavenport See my comment to Paul above. Maybe would you help me run the code in order to test it on submissions?

src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -112,10 +112,18 @@ public Grader(String repoUrl, String netId, Observer observer, Phase phase) thro
...
         this.observer = observer;
     }
+    private GradingSettings getGradingSettings() {
+        // FIXME! Import from some dynamic location
+        return new GradingSettings(
+                5,
+                10,
+                10
+        );
+    }
```
pawlh, line 126: I like the future prep here 💯

frozenfrank:
That's fine, I'll tackle it. Thanks for letting me know.

frozenfrank:
The underlying issue has already been resolved by #263.

If we want to preserve any of the other ideas from this PR, they'll come from a new one.

PR #177, opened 2024-02-21
Rubricify canvas

19mdavenport:
Adds functionality for sending rubrics to canvas

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/Grader.java
+++ b/src/main/java/edu/byu/cs/autograder/Grader.java
@@ -281,7 +281,8 @@ public abstract class Grader implements Runnable {
         try {
             CanvasIntegration.submitGrade(userId, assignmentNum, scores, comments, submission.notes());
         } catch (CanvasException e) {
-            LOGGER.error("Error submitting score for user " + submission.netId(), e);
+            LOGGER.error("Error submitting to canvas for user " + submission.netId(), e);
+            throw new RuntimeException("Error contacting canvas to record scores");
         }
 
     }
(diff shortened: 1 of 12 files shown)


src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -131,16 +132,26 @@ public void run() {
...
-            Rubric.Results passoffResults = runTests();
-            Rubric.Results customTestsResults = runCustomTests();
+            Rubric.Results passoffResults = null;
+            if(rubricConfig.passoffTests() != null) {
+                passoffResults = runTests();
+            }
+            Rubric.Results customTestsResults = null;
+            if(rubricConfig.unitTests() != null) {
+                customTestsResults = runCustomTests();
+            }
```
pawlh, line 149: thanks for adding these guards

src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -228,26 +239,47 @@ private Submission saveResults(Rubric rubric, int numCommits) {
...
+            scores.put(id, results.score() * lateAdjustment);
+            comments.put(id, results.notes());
+        }
+
 
         try {
-            CanvasIntegration.submitGrade(userId, assignmentNum, score, submission.notes());
+            CanvasIntegration.submitGrade(userId, assignmentNum, scores, comments, submission.notes());
         } catch (CanvasException e) {
             LOGGER.error("Error submitting score for user " + submission.netId(), e);
```
pawlh, line 284: I know we weren't already doing this before your change, but I think we should throw an exception here instead so that the user would be notified that they didn't  get a grade in Canvas

pawlh: Instead of just logging it

PR #189, opened 2024-02-23
Backend move phases to resources

19mdavenport:
This moves the phases folder to inside the resources folder and then the code can pull it out when the server starts


pawlh:
@19mdavenport Would you mind resolving the conflicts here the next time you get the chance? I'd like to start thinking about automatic deployment again pretty soon

pawlh (approved):

PR #198, opened 2024-02-29
124 stop grading if it lowers score

DallinFromEarth:
The grade now runs, then checks the score it calculated against the score stored in canvas. If the new score is equal to or lower than the score in canvas, it doesn't save it to canvas, and it notifies the user.

Issue #124, opened 2024-01-25
backend: stop grade sending to canvas if the late penalty results in a lower grade


DallinFromEarth:
If there is a case where sending the grade to canvas lowers the grade, don't send it and notify the student

DallinFromEarth:
Front end issue to solve at the same time: a student ran his code, the extra credit passed, but the required tests didn't, and now he can't resubmit. Make sure that the checker looks at number of required tests failed

frozenfrank:
This is reopened by #204

DallinFromEarth:
completed by #118 and #206

PR #204, opened 2024-03-04
Revert "124 stop grading if it lowers score"

pawlh:
Reverts softwareconstruction240/autograder#198


frozenfrank:
This reopend #124

PR #205, opened 2024-03-05
Update docker-compose.yml app depends on db

19mdavenport:


pawlh:
This does not resolve the db initialization error

PR #206, opened 2024-03-05
124 stop grading if it lowers score - fix null exception

DallinFromEarth:
This pull request is to prevent students from saving the score of a submission if it is lower than their current score in canvas

The earlier version of this addition #198 had an error that we failed to catch in testing where it would crash if the student had no score in canvas. This has now been addressed by simply saying a score is 0 if the canvas API returns a record with a score value of null


19mdavenport (dismissed):
This pull request is missing the code that actually checks the scores from the previous pull request #198. This might be a merge conflict in grader that got rid of it, or just missing in this version of this branch

DallinFromEarth:
@19mdavenport interesting, the code is in the branch itself but doesn't seem to be part of the PR. I'll have to look into it tomorrow.

19mdavenport:
I cloned the branch and it wasn't there for me. Could it have been the merge from main that did that?

DallinFromEarth:
Interesting, it seems that was the case. I reverted that commit and it looks like the code is back

DallinFromEarth:
When we're testing this pr, make sure you clear the test student to make sure the problem this originally created is fixed

DallinFromEarth:
fix saveResults() so that it doesn't send it to canvas. Apparently this didn't get changed back

pawlh (dismissed):
It looks like the issues have been resolved

19mdavenport:
The nullability issue has been resolved, but it is now sending to canvas twice on successful submissions and once on unsuccessful submissions
image

pawlh:
> fix saveResults() so that it doesn't send it to canvas. Apparently this didn't get changed back

> The nullability issue has been resolved, but it is now sending to canvas twice on successful submissions and once on unsuccessful submissions

Lol I forgot to actually check Canvas. Good catch both of you

pawlh (approved):

pawlh:
> I believe this is fixed

I also think it is. I am going to merge this tonight, so hopefully if there are any issues we'll know before mass submissions start later in the day

PR #207, opened 2024-03-06
Adjust quality grading (spring term)

19mdavenport:
This changes the quality grading categories to be all or nothing rather than each check being a certain number of points

### DIFF
--- a/phases/libs/qualityRubric.json
+++ b/phases/libs/qualityRubric.json
@@ -37,7 +37,8 @@
       "value": 30,
       "reporters": [
         "[UnusedMethodReporter]",
-        "[CommentedCode]"
+        "[CommentedCode]",
+        "[NoCodeInFile]"
       ]
     },
     {
--- a/phases/libs/qualityRubric.json
+++ b/phases/libs/qualityRubric.json
@@ -4,48 +4,48 @@
       "name": "Naming",
       "value": 25,
       "reporters": [
-        "[PackageName]",
-        "[TypeName]",
-        "[MethodName]",
-        "[ConstantName]",
-        "[MemberName]",
-        "[LocalVariableName]",
-        "[RecordComponentName]",
-        "[ParameterName]",
-        "[LambdaParameterName]",
-        "[StaticVariableName]",
-        "[LocalFinalVariableName]",
-        "[PatternVariableName]",
-        "[RecordTypeParameterName]",
-        "[MethodTypeParameterName]",
-        "[ClassTypeParameterName]",
-        "[InterfaceTypeParameterName]"
+        "PackageName",
+        "TypeName",
+        "MethodName",
+        "ConstantName",
+        "MemberName",
+        "LocalVariableName",
+        "RecordComponentName",
+        "ParameterName",
+        "LambdaParameterName",
+        "StaticVariableName",
+        "LocalFinalVariableName",
+        "PatternVariableName",
+        "RecordTypeParameterName",
+        "MethodTypeParameterName",
+        "ClassTypeParameterName",
+        "InterfaceTypeParameterName"
       ]
     },
     {
       "name": "Code Decomposition",
       "value": 30,
       "reporters": [
-        "[DuplicateLines]",
-        "[DuplicateBlock]",
-        "[MethodLength]",
-        "[FileLength]"
+        "DuplicateLines",
+        "DuplicateBlock",
+        "MethodLength",
+        "FileLength"
       ]
     },
     {
       "name": "Code Readability",
       "value": 30,
       "reporters": [
-        "[UnusedMethodReporter]",
-        "[CommentedCode]",
-        "[NoCodeInFile]"
+        "UnusedMethodReporter",
+        "CommentedCode",
+        "NoCodeInFile"
       ]
     },
     {
       "name": "Package Structure",
       "value": 15,
       "reporters": [
-        "[FilesPerPackage]"
+        "FilesPerPackage"
       ]
     }
   ]
--- a/phases/libs/qualityRubric.json
+++ b/phases/libs/qualityRubric.json
@@ -2,163 +2,49 @@
   "categories": [
     {
       "name": "Naming",
-      "value": 7.5,
-      "items": [
-        {
-          "reporters": [
-            {
-              "name": "[PackageName]",
-              "weight": 1
-            }
-          ],
-          "weight": 0
-        },
-        {
-          "reporters": [
-            {
-              "name": "[TypeName]",
-              "weight": 1
-            }
-          ],
-          "weight": 3
-        },
-        {
-          "reporters": [
-            {
-              "name": "[MethodName]",
-              "weight": 1
-            }
-          ],
-          "weight": 3
-        },
-        {
-          "reporters": [
-            {
-              "name": "[ConstantName]",
-              "weight": 0
-            },
-            {
-              "name": "[MemberName]",
-              "weight": 3
-            },
-            {
-              "name": "[LocalVariableName]",
-              "weight": 3
-            },
-            {
-              "name": "[RecordComponentName]",
-              "weight": 3
-            },
-            {
-              "name": "[ParameterName]",
-              "weight": 3
-            },
-            {
-              "name": "[LambdaParameterName]",
-              "weight": 2
-            },
-            {
-              "name": "[StaticVariableName]",
-              "weight": 2
-            },
-            {
-              "name": "[LocalFinalVariableName]",
-              "weight": 1
-            },
-            {
-              "name": "[PatternVariableName]",
-              "weight": 1
-            }
-          ],
-          "weight": 5
-        },
-        {
-          "reporters": [
-            {
-              "name": "[RecordTypeParameterName]",
-              "weight": 1
-            },
-            {
-              "name": "[MethodTypeParameterName]",
-              "weight": 1
-            },
-            {
-              "name": "[ClassTypeParameterName]",
-              "weight": 1
-            },
-            {
-              "name": "[InterfaceTypeParameterName]",
-              "weight": 1
-            }
-          ],
-          "weight": 1
-        }
+      "value": 25,
+      "reporters": [
+        "[PackageName]",
+        "[TypeName]",
+        "[MethodName]",
+        "[ConstantName]",
+        "[MemberName]",
+        "[LocalVariableName]",
+        "[RecordComponentName]",
+        "[ParameterName]",
+        "[LambdaParameterName]",
+        "[StaticVariableName]",
+        "[LocalFinalVariableName]",
+        "[PatternVariableName]",
+        "[RecordTypeParameterName]",
+        "[MethodTypeParameterName]",
+        "[ClassTypeParameterName]",
+        "[InterfaceTypeParameterName]"
       ]
     },
     {
       "name": "Code Decomposition",
-      "value": 9.0,
-      "items": [
-        {
-          "reporters": [
-            {
-              "name": "[DuplicateLines]",
-              "weight": 1
-            },
-            {
-              "name": "[DuplicateBlock]",
-              "weight": 1
-            }
-          ],
-          "weight": 3
-        },
-        {
-          "reporters": [
-            {
-              "name": "[MethodLength]",
-              "weight": 1
-            },
-            {
-              "name": "[FileLength]",
-              "weight": 1
-            }
-          ],
-          "weight": 2
-        }
+      "value": 30,
+      "reporters": [
+        "[DuplicateLines]",
+        "[DuplicateBlock]",
+        "[MethodLength]",
+        "[FileLength]"
       ]
     },
     {
       "name": "Code Readability",
-      "value": 9.0,
-      "items": [
-        {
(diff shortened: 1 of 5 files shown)


phases/libs/qualityRubric.json
```diff
@@ -2,163 +2,50 @@
...
-            {
-              "name": "[InterfaceTypeParameterName]",
-              "weight": 1
-            }
-          ],
-          "weight": 1
-        }
+      "value": 25,
+      "reporters": [
+        "[PackageName]",
```
frozenfrank, line 7: I really like the grouping of multiple `reporters` into a single `category`. What is the purpose of the square brackets around each of the names? Is that part of the string matching, or part of the official name of the rule? If it's not _part of the name_, could we remove them and make it part of the processing phase?

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -39,17 +41,17 @@ public QualityOutput runQualityChecks(File stageRepo) {
         output = output.replaceAll(stageRepo.getAbsolutePath(), "");
         output = output.replaceAll(stageRepo.getPath(), "");
 
-        QualityAnalysis analysis = parseAnalysis(output);
-        float score = getScore(analysis);
-        String results = getResults(analysis);
-        String notes = getNotes(analysis);
-        return new QualityOutput(score, results, notes);
+        parseOutput(output);
```
frozenfrank, line 44: This section of code seems much cleaner and more readable than before. Looking good!

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -64,82 +66,68 @@ private QualityAnalysis parseAnalysis(String output) {
...
-                                .forEach(s -> reporterResultsBuilder.append("\t\t").append(s).append("\n"));
-                    }
-                    if (!reporterResultsBuilder.isEmpty()) {
-                        categoryResultsBuilder.append("\t").append(reporter.name()).append(":\n")
-                                .append(reporterResultsBuilder);
-                    }
+            for(String reporter : category.reporters()) {
+                StringBuilder reporterResultsBuilder = new StringBuilder();
+                if (errors.containsKey(reporter)) {
+                    errors.get(reporter).forEach(s -> reporterResultsBuilder.append("\t\t").append(s).append("\n"));
```
frozenfrank, line 88: I'm sure this code _works_ just perfectly, and I can tell that it's doing some kind of string concatenation to build a resulting string that will be displayed to the user. 

It's kind of hard for me to read what kind of string this code is supposed to be building. Sometimes comments are problematic because they grow stale, but I wonder if it would be helpful here to have an example string of the output so I can follow along in the code...

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -64,82 +66,68 @@ private QualityAnalysis parseAnalysis(String output) {
...
 
 
-    public record QualityOutput(float score, String results, String notes) {}
-
-    private record QualityAnalysis(Map<String, List<String>> errors, List<String> warnings) {}
-
-    private record QualityRubricReporter(String name, int weight) {}
-
-    private record QualityRubricItem(Set<QualityRubricReporter> reporters, int weight) {}
+    public record QualityAnalysis(float score, String results, String notes) {}
```
frozenfrank, line 128: Would this be an appropriate place to use JavaDoc comments to describe these parameters? The `score` value is pretty clear, but I'm actually not sure about the difference between `results` and `notes`. Some defining documentation could help prevent confusion in the future.

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -39,17 +41,17 @@ public QualityOutput runQualityChecks(File stageRepo) {
...
+        String notes = getNotes();
+        return new QualityAnalysis(score, results, notes);
     }
 
 
-    private QualityAnalysis parseAnalysis(String output) {
-        Map<String, List<String>> errors = new HashMap<>();
-        List<String> warnings = new ArrayList<>();
+    private void parseOutput(String output) {
+        errors = new HashMap<>();
```
frozenfrank, line 53: Some people will argue that the code is cleaner when each method is architected as a [pure function](https://en.wikipedia.org/wiki/Pure_function), rather than relying on and mutating local variables.

In this case, it would seem straight forward to have this `parseOutput()` method return the `HashMap<> errors` and the `ArrayList<> warnings`, maybe as a record or a pair, and then explicitly pass them in to the subsequent methods that depend on them. The main advantage of this approach is that each method individually behaves the same given the same input every time. As a general rule, it's safer to follow acknowledging that this code seems to run just fine as it is.

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -64,82 +66,68 @@ private QualityAnalysis parseAnalysis(String output) {
...
-                if (analysis.errors().containsKey(reporter.name())) {
-                    deductions += weight;
-                }
-            }
+    private boolean categoryPasses(QualityRubricCategory category) {
+        for(String reporter : category.reporters()) {
+            if(errors.containsKey(reporter)) return false;
         }
-        return category.value() * (1 - (deductions / totalWeight));
+        return true;
```
frozenfrank, line 124: I love how simple this logic becomes with the simplified grading!

src/main/resources/frontend/src/views/PhaseView/RubricTable.vue
```diff
@@ -26,51 +23,9 @@ const failedTests = (testResults: TestResult) => {
...
-        <td>{{ Math.round(rubric.quality.results.score) }} / {{ rubric.quality.results.possiblePoints }}</td>
-        <td
-            class="selectable"
-            @click="() => {
-              selectedTestResults = undefined
-              selectedTextResults = rubric.quality.results.textResults;
-            }"
-        >Click here</td>
-      </tr>
+      <RubricTableRow :rubric-item="rubric.passoffTests" :results-clicked="() => {selectedTestResults = rubric.passoffTests.results.testResults;}"/>
```
frozenfrank, line 26: I love this use of special components designed for a given purpose!

src/main/resources/frontend/src/views/PhaseView/RubricTableRow.vue
```diff
@@ -0,0 +1,51 @@
...
+  rubricItem: RubricItem;
+  resultsClicked: () => void;
+}>();
+
+const failedTests = (testResults: TestResult) => {
+  return testResults && (testResults.numTestsFailed > 0 || testResults.numTestsPassed === 0);
+}
+
+const addColor = (qualityNotes: string) => {
+  return qualityNotes.replace(/✓/g, `<span class="success">✓</span>`).replace(/✗/g, `<span class="failure">✗</span>`);
```
frozenfrank, line 14: Note: Javascript offers an [String.replaceAll()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) method which may be helpful here.

frozenfrank (commented):
Overall, the approach looks good, and I'm sure it works just great. I left some comments about techniques that could be employed to improve the readability of the code and the long-term maintainability of the files. 

The changes already proposed in the PR are clearly better for the students. Thank you for building such powerful and useful systems @19mdavenport

phases/libs/qualityRubric.json
19mdavenport: That is how it's spit out by checkstyle, but I'll try to put it into the processing instead

19mdavenport: How's [this?](https://github.com/softwareconstruction240/autograder/pull/207/commits/06e599fd57778adac33c951f6bbdf0d1164dc8cf)

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
19mdavenport: Example output:
Code Decomposition:
	DuplicateBlock:
		[ERROR] /server/src/main/java/service/Service.java:19:34: Duplicated code from /server/src/main/java/server/Server.java:170 [DuplicateBlock]
Code Readability:
	UnusedMethodReporter:
		[ERROR] /shared/src/main/java/model/GameData.java:30: Unused method clearBoard [UnusedMethodReporter]
		[ERROR] /client/src/main/java/ui/ChessClient.java:75: Unused method quit [UnusedMethodReporter]
		[ERROR] /client/src/main/java/ui/ChessClient.java:179: Unused method legal [UnusedMethodReporter]
		[ERROR] /client/src/main/java/ui/ChessClient.java:122: Unused method list [UnusedMethodReporter]
		[ERROR] /client/src/main/java/ui/ChessClient.java:170: Unused method redraw [UnusedMethodReporter]

src/main/resources/frontend/src/views/PhaseView/RubricTableRow.vue
19mdavenport: IntelliJ gives an error when using replaceAll, which is why I opted for just using replace. Paul said he'd rather have old javascript than have errors

pawlh: `replaceAll` was added with ES2021 (https://github.com/daumann/ECMAScript-new-features-list/blob/main/ES2021.MD), and for some reason our TypeScript configuration is running against an older specification.

I looked into this before, but couldn't get TypeScript to run against the newer version. I'm sure it's possible, so if someone knows how to fix that I'd be all for it.

frozenfrank: Here is the [Vue documentation](https://v2.vuejs.org/v2/guide/typescript) on using Typescript.

It looks like the current app hasn't had that configuration, but I'll propose some changes to see if they solve the problem

frozenfrank: This issue should be resolved with #223. After merging that PR into this branch, we should be allowed to use `.replaceAll()` without complaints.

phases/libs/qualityRubric.json
frozenfrank: I love it! Thanks 😄

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
frozenfrank: That's helpful to have in the PR. I suppose anyone working on the code will just need to reference the working product. It's probably safer to do it that way then risk having stale comments in the codebase.

Looks good to me

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -66,67 +74,100 @@ private void parseOutput(String output) {
...
         }
+        return new QualityOutput(errors, warnings);
     }
 
-    private float getScore() {
+    /**
+     * Evaluates the score for a quality analysis
+     *
+     * @param output errors and warnings from quality analysis
+     * @return the score as a float from 0 to 1
```
frozenfrank, line 84: Ooh good point. Knowing the range of the value is really helpful. I nearly forgot that would be a tricky point moving forward!

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
```diff
@@ -66,67 +74,100 @@ private void parseOutput(String output) {
...
-
-    private boolean categoryPasses(QualityRubricCategory category) {
-        for(String reporter : category.reporters()) {
-            if(errors.containsKey(reporter)) return false;
+    /**
+     * Determines if a particular category had any quality errors
+     *
+     * @param category category to check
+     * @param output   errors and warnings from quality analysis
+     * @return false if an error was present for the category, false otherwise
```
frozenfrank, line 151: Did you mean to say `... true otherwise` at the end of this line?

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
frozenfrank: The new documentation is most helpful!

frozenfrank (approved):
I think this is really good. Thanks for making adjustments based on my feedback from earlier!

I still hope that we can switch to `String.replaceAll()`, but I see the reasoning for obeying the compiler warnings. I did submit a PR into this branch that hopefully resolves the issue; it doesn't need to be part of this PR. GitHub will automatically switch it to point to the `main` branch if this is merged, and we can continue the conversation longer.

Thanks for making these adjustments!

src/main/java/edu/byu/cs/autograder/QualityAnalyzer.java
19mdavenport: I did. Thank you.

frozenfrank: Lol we both committed to resolve this issue at the same time 😆 You pushed before I did though @19mdavenport...

image

Issue #158, opened 2024-02-09
frontend/admin: make submissions tab searchable and sortable


DallinFromEarth:
look into using AG-Grid

PR #213, opened 2024-03-09
212 admin add panel for student information

DallinFromEarth:
There is now a new tab: "students"

This tab shows a list of all students from the server database, their netID, and a clickable Github URL. Clicking on either the student's name or netID will open up a pop-up window with all of their submissions. (Which you can then open another pop-up in)


DallinFromEarth:
branch #158 has taken the new table library used for the student tab in this pr, and applied it to the submissions tab. Theres still a few small front-end things to update there, but that can be pulled into this pr if it makes more sense.

19mdavenport (commented):
I don't want to approve this for merging without @pawlh seeing this but this looks pretty good to me 👍

19mdavenport (dismissed):
Is there a yarn dependency that needs to be added for ag-grid-vue3?

pawlh (approved):
❤️

PR #219, opened 2024-03-11
172 holidays and weekends dont extend late calculation

### LINKED ISSUE #172: backend: Holidays and weekends don't unnecessarily extend the late penalty calculation
## Description
The due date is the last opportunity to not have any late penalties applied, but if the day after it is a holiday or a weekend, then turning in the assignment two calendar days later should only count as **one day late**.

After the due date, the assignment is late by 1 day, but it won't go up to 2 until the next non-holiday, non-weekend is over

## Worklist
This will require:
* Defining a list of recognized holidays
* Allowing a list of recognized holidays to be stored in `GradingSettings`
* Adapting the `daysLate` calculation to skip holidays

The final decision point for these calculations will be in `src/main/java/edu/byu/cs/autograder/Grader.java`.

frozenfrank:
This resolves #172 with a very robust capability of reading configuration strings. This does not set up the dynamic string configuration since it needs to be decided on a higher level. That issue is contemplated in #156.

This introduces the ability to respond to configuration values in two different formats:
* **single line strings**, optimized for storing in a cell of a configuration table which can be updated
* **multi-line strings**, optimized for storing in a file with readable comments and understandable language

Examples of both formats are tested in appropriate test files.

While this isn't primarily concerned with this, we do introduce hard-coded settings to carry us through the 2024 calendar year. Some dates may be incorrect; rather than adjusting them here, we should choose a method for storing them more dynamically which this PR supports. 

Credit to @prattnj for building the original pattern from which we have added this robust functionality. Thanks to @19mdavenport for support in creating the appropriate test files and determine the correct behavior.


pawlh:
Woops, sorry I did not intend to add `switch back to cache action` to your PR

frozenfrank:
> Woops, sorry I did not intend to add switch back to cache action to your PR, although there shouldn't be a problem if it stays here

@pawlh That's ok. Since I see you already have it on another branch as well, I'll remove it.

I am going to add one more test before I unmark this as a "draft." I forgot to validate that the due dates actually extend properly when the holidays are configured!

frozenfrank:
@leesjensen You may appreciate thoroughness of the testing of this method. Because of the test cases evaluated, I am now very confident that this works as it is supposed to work. Hopefully, this will be appreciated as a way of future proofing this code and making it more accessible to future TA's.

frozenfrank (commented):
I feel like I did a great job :)

frozenfrank:
For good measure, a screenshot of all the tests passing:

image

pawlh (approved):

PR #226, opened 2024-03-12
Improve error output

19mdavenport:
This displays additional error messages in test output of anything printed out to standard error (other than spark initialization) during testing


19mdavenport:
@pawlh I would appreciate it if you could take a look at output provided by this at some point

pawlh (approved):

Issue #209, opened 2024-03-06
Students Receive Error Messages Relating to Temp Files


frozenfrank:
## Overview
It seems that the autograder is sometimes throwing errors at the students related to compiling information in the temp folder. While I'm not sure what's causing these problems, it does interfere with the student's abilities to submit their code.

## Screenshots
In both of the examples below, I saw the autograder throw these issues, but then I also worked with the student to pass them off. Their code seemed just fine in both cases.

### Student 1
While submitting Phase 3, the autograder threw errors about not being able to run a program and mentioned her temp directory. I passed her off manually and the tests worked just fine locally. Occurred on 27 Feb 2024.
![image](https://github.com/softwareconstruction240/autograder/assets/19808440/ccd300dd-03dd-4ded-8ae1-ac5c0ceb72b3)

### Student 2. 
While submitting Phase 1, the autograder threw errors about a `db.properties` file. Unfortunately, she didn't realize that this wasn't her issue very quickly. Occurred on 5 Mar 2024.
![image](https://github.com/softwareconstruction240/autograder/assets/19808440/76c29423-f78c-418d-9c72-b6bc3aedc9cc)

frozenfrank:
I also haven't seen any problems near the end of the semester. We'll mark this one closed by some combination of the many adjustments being made earlier.

PR #232, opened 2024-03-13
229 backend verify that tests exist before attempting to compile

19mdavenport:
This commit requires tests exist before compiling (having no tests just means getting a 0 for that rubric item). This also requires phase 3 tests be in the serviceTests folder


pawlh (approved):
💯

frozenfrank:
This issue at least partially resolves #209, but I'm not sure if that issue is totally resolved. There could be other places where files are accessed without verifying that they exist which could still lead to that issue.

PR #233, opened 2024-03-13
160 hold off submitting scores for low commit submissions

frozenfrank:
This code should receive and accept a low-commit submission from a student, and then continue processing the submission. It should report the low-commit submission, not report the grade to canvas, and alert the student to talk to a TA to receive credit.

Resolves #160, resolves #230.

This includes work from #284.

## Database modification
Run the following SQL statements to add new columns to the database
```sql
ALTER TABLE submission ADD `verified_status` VARCHAR(30);
ALTER TABLE submission ADD `verification` JSON;
ALTER TABLE submission DROP COLUMN `num_commits`;
ALTER TABLE submission DROP COLUMN `results`;
```

## TODO
- [ ] Verify that new submissions any applied penalty percentage from an earlier deduction.

PR #234, opened 2024-03-13
Add ability to send a canvas comment without a grade

19mdavenport:
This adds the ability to submit to canvas on an assignment without overriding the grade, only adding a comment
@frozenfrank requested this feature for #233


frozenfrank (approved):
This code looks good to me! Thanks for responding to requests for improvements so quickly and skillfully. You're a master @19mdavenport :)

FYI I added a commit that explicitly documents the nullable-ness of the parameters.

PR #238, opened 2024-03-13
Load only 25 Submissions initially

DallinFromEarth:
The Submissions View now only loads the most recent 25 latest submissions. (Latest submission meaning the most recent submission per student per phase). There is now a button at the bottom to load in all latest submissions.

This was done by adding an optional batchSize parameter to the getAllLatestSubmissions() on the DAO that uses a LIMIT statement as suggested by @frozenfrank. 

The specific amount of submissions that is initially loaded is determined by a const in the submissions view, so is easily adjustable.

### DIFF
--- a/src/main/java/edu/byu/cs/dataAccess/SubmissionDao.java
+++ b/src/main/java/edu/byu/cs/dataAccess/SubmissionDao.java
@@ -32,11 +32,19 @@ public interface SubmissionDao {
     Collection<Submission> getSubmissionsForUser(String netId);
 
     /**
-     * Gets all submissions for the given phase
+     * Gets all latest submissions
      *
-     * @return all submissions for the given phase
+     * @return all latest submissions
      */
     Collection<Submission> getAllLatestSubmissions();
+
+    /**
+     * Gets the X most recent latest submissions
+     *
+     * @param batchSize defines how many submissions to return. Set batchSize to -1 to get All submissions
+     * @return the most recent X submissions
+     */
+    Collection<Submission> getAllLatestSubmissions(int batchSize);
 
     /**
      * Removes all submissions for the given netId
--- a/src/main/java/edu/byu/cs/dataAccess/SubmissionDao.java
+++ b/src/main/java/edu/byu/cs/dataAccess/SubmissionDao.java
@@ -41,7 +41,7 @@ public interface SubmissionDao {
     /**
      * Gets the X most recent latest submissions
      *
-     * @param batchSize defines how many submissions to return. Set batchSize to -1 to get All submissions
+     * @param batchSize defines how many submissions to return. Set batchSize to a negative int to get All submissions
      * @return the most recent X submissions
      */
     Collection<Submission> getAllLatestSubmissions(int batchSize);
--- a/src/main/java/edu/byu/cs/dataAccess/memory/SubmissionMemoryDao.java
+++ b/src/main/java/edu/byu/cs/dataAccess/memory/SubmissionMemoryDao.java
@@ -4,7 +4,6 @@ import edu.byu.cs.dataAccess.SubmissionDao;
 import edu.byu.cs.model.Phase;
 import edu.byu.cs.model.Submission;
 
-import java.time.Instant;
 import java.util.Collection;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -57,7 +56,7 @@ public class SubmissionMemoryDao implements SubmissionDao {
                 }
             });
 
-            batchSize += -1;
+            batchSize -= 1;
             if (batchSize == 0) { break; }
         }
 
--- a/src/main/java/edu/byu/cs/dataAccess/memory/SubmissionMemoryDao.java
+++ b/src/main/java/edu/byu/cs/dataAccess/memory/SubmissionMemoryDao.java
@@ -4,6 +4,7 @@ import edu.byu.cs.dataAccess.SubmissionDao;
 import edu.byu.cs.model.Phase;
 import edu.byu.cs.model.Submission;
 
+import java.time.Instant;
 import java.util.Collection;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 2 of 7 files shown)


src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
```diff
@@ -20,25 +20,36 @@ import {nameFromNetId} from "@/utils/utils";
 
 const selectedSubmission = ref<Submission | null>(null);
 const selectedStudent = ref<User | null>(null);
+const DEFAULT_SUBMISSIONS_TO_LOAD = 25;
```
frozenfrank, line 23: We need a client side config system similar to #156.

src/main/resources/frontend/src/services/adminService.ts
```diff
@@ -66,9 +66,10 @@ export const userPatch = async (user: UserPatch)=> {
     }
 }
 
-export const submissionsLatestGet = async (): Promise<Submission[]> => {
+export const submissionsLatestGet = async (batchSize?: number): Promise<Submission[]> => {
```
frozenfrank, line 69: From a defensive perspective, I would argue that the default value behavior of this method should request a finite number of submissions. Then, a developer must specifically ask for all submissions in order to get them.

It's tricky because any value we put as the default will be arbitrary, and we don't have a good system in place to provide configurations. Ideally, we would change the paradigm of this method, but maybe it's not reasonable to do until we have other supporting systems in place.

src/main/java/edu/byu/cs/dataAccess/SubmissionDao.java
```diff
@@ -32,12 +32,20 @@ public interface SubmissionDao {
     Collection<Submission> getSubmissionsForUser(String netId);
 
     /**
-     * Gets all submissions for the given phase
+     * Gets all latest submissions
      *
-     * @return all submissions for the given phase
+     * @return all latest submissions
```
frozenfrank, line 37: Note that these API specs do not specify which order the submissions should be returned in. Can we expect them to be in ascending or descending order, or could they come to us in any order? It would seem reasonable to define that behavior as part of this API, and if so, the documentation should reflect the decision.

src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
```diff
@@ -20,25 +20,36 @@ import {nameFromNetId} from "@/utils/utils";
...
+let allSubmissionsLoaded = false;
 
 onMounted(async () => {
-  const submissionsData = await submissionsLatestGet();
-  submissionsData.sort((a, b) => b.timestamp.localeCompare(a.timestamp)) // Sort by timestamp descending
+  const submissionsData = await submissionsLatestGet(DEFAULT_SUBMISSIONS_TO_LOAD);
+  //submissionsData.sort((a, b) => b.timestamp.localeCompare(a.timestamp)) // Sort by timestamp descending
```
frozenfrank, line 28: It seems this line has been removed, and you are relying on the server returning sorted results to you.

## Trusting Results to be Pre Sorted
Per the comment above, that isn't strictly defined which makes it an unsafe assumption for now. Breaking the fourth wall to go look at the server code from the client, it seems that the SQLSubmissionDAO uses an `ORDER BY` clause which will do the behavior, but it seems that the MemorySubmissionDAO **does not** manage the order of the returned results.

It's not safe to trust that the results will come to you properly until it's defined in the API. At that point, we can point fingers at the server if things break, but honestly, the client should be responsible for sorting the results because they'll usually want diverse kinds of sorts.

## Leaving the comment behind
If after discussing all the reasons, it's decided that this line isn't needed, the line should be removed entirely. We can always retrieve the code again from the `git` file history. There's no need to leave commented out code in the file in case you need it again. Besides, we would dock our students for this behavior (I hope)!

frozenfrank (changes requested):
This will be a helpful PR!

I've left a few comments about adjustments that would be smart to make or clarify before we merge the code in.

Thanks for building all these cool features for us @DallinFromEarth 🚀

src/main/java/edu/byu/cs/dataAccess/memory/SubmissionMemoryDao.java
```diff
@@ -48,7 +56,11 @@ public Collection<Submission> getAllLatestSubmissions() {
                     return v;
                 }
             });
-        });
+
+            batchSize += -1;
```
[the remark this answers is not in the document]
DallinFromEarth: 🤦

src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
```diff
@@ -88,7 +93,12 @@ SELECT MAX(timestamp)
                                 FROM submission
                                 GROUP BY net_id, phase
                             )
-                            """);
+                            ORDER BY timestamp DESC
+                            """ +
+                            (batchSize == -1 ? "" : "LIMIT ?"));
```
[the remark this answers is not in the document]
DallinFromEarth: I agree that would be a good idea. I'll go ahead and change it so that any int below zero returns all latest submissions

src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
DallinFromEarth: I agree, though at this point I'm not sure if it's worth it to set this up in this PR.

src/main/resources/frontend/src/services/adminService.ts
DallinFromEarth: That seems to be the best way to do it from a defensive perspective. However, I'm not sure how to define a default amount without setting up the client-side-settings you mentioned in another comment

src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
DallinFromEarth: The comment should be removed, and I will do so promptly. It was left there while I was making sure that it was still sorted after refactoring some things. 

The commented code is not needed, AND we are not trusting the results to be pre-sorted. The table automatically sorts the data descending based on the timestamp column.

src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
frozenfrank: Agreed. Not worth setting up in this PR, but the reference to the other issue will show up in the history and maybe help us remember where we we wanted to use it when we do implement it.

src/main/java/edu/byu/cs/dataAccess/SubmissionDao.java
frozenfrank: Oh great! I didn't see the client side code that was sorting it in the table. Knowing that it's already taken care of in that context resolves all my concerns. 

Great work delivering really cool and well crafted solutions, Dallin!

PR #243, opened 2024-03-15
Improve Honor Checker

prattnj:
## What's New
- Test Student is no longer included in the Honor Checker compilation results

## Improvements
- The resulting zip file now contains a zip file for each student rather than a directory
   - This is necessary for the Honor Checker and it makes extracting the top-level zip about 500x faster

## Bug Fix
In its previous state, the resulting zip file for each section was never removed from the root directory. Now it is (existing ones, which there's at least one of, will have to be manually deleted).


frozenfrank (approved):
Without running the code myself, the code changes line up with the description of the PR. They're each well done and appear to be the correct change.

Good adjustments!

pawlh:
Looks good, but you should probably also change the frontend note about this too

https://github.com/softwareconstruction240/autograder/blob/5af03488beda0bb91fc9d2a940d5b4346c51271a/src/main/resources/frontend/src/views/AdminView/HonorChecker.vue#L48

PR #253, opened 2024-03-20
Make logs accessible

19mdavenport:


src/main/java/edu/byu/cs/server/Server.java
```diff
@@ -65,17 +67,25 @@ public static void setupEndpoints() {
 
                 patch("/user/:netId", userPatch);
 
-                get("/submissions/latest", latestSubmissionsGet);
+                path("/submissions", () -> {
```
DallinFromEarth, line 70: Big fan of condensing these endpoints

DallinFromEarth (approved):
Looks good to me! Nice simple approach that gets the job done! Runs perfectly

pawlh (approved):
Eventually I'd like to see a more comprehensive logging solution (#48 is a WIP), but this is a great short term solution, so let's merge it

PR #256, opened 2024-03-20
Allow submissions from admin panel

DallinFromEarth:
## Main Update
The submissions tab now features a way to run any GitHub repo on any phase as an admin. It runs as the admin and is saved like any other submission, just not saved anywhere in Canvas. Admin submissions also bypass checks on number of commits and whether that commit has already been graded.

### Technical Details
- A `Submission` now includes a `admin` field, which if true, indicates that the submission was run by an admin.
  - This has been updated in model objects and in the SQL database
- The submission runs and is saved to the database under the admin's `netId`. You can access all their submissions by clicking on their name from the submissions tab, but they still do not appear on the students tab. This was an intentional choice, that can be changed if we want it to be different.
- A new endpoint has been added for these admin submissions at `/api/admin/submit` as a POST method
- The endpoint for student submissions has been refactored to be more readable and more modular. This allowed code to be reused for the new admin endpoint
- Theres also a fun new `Dropdown` element that I made for the new admin phase selection, but its written so it can be reused anywhere else we need it

### Deployment and Development Note:
Because of the new `admin` field in the submissions table, this needs to be updated in the development database, as well as any local databases being used by the team for development. This is the SQL command I ran from my console that should update the tables as necessary:
```
ALTER TABLE submission
    ADD COLUMN admin BOOLEAN DEFAULT FALSE;
```

### **HELP WANTED**
Because this PR touches the end point for the grader, it would be greatly appreciated for anyone to download this PR and verify that the grader runs correctly from the new admin view AND the student view.


src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
19mdavenport, line 1: The create tables statement doesn't include the admin field

19mdavenport (changes requested):

src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
DallinFromEarth: I'm looking at this file `SubmissionSqlDao.java` and don't see a create tables statement at all. Theres an alter statement in the PR that you can run from the console to get the new field

19mdavenport: src/main/java/edu/byu/cs/dataAccess/sql/SqlDb.java

pawlh (commented):
Looks good!

I am going to hold off merging this until tomorrow when there are fewer submissions coming in

src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
```diff
@@ -68,10 +85,41 @@ const columnDefs = reactive([
...
+      <template v-slot:dropdown-parent>
+        <button>Grade Repo</button>
+      </template>
+      <template v-slot:dropdown-items>
+        <a @click="adminSubmit('0')">Phase 0</a>
+        <a @click="adminSubmit('1')">Phase 1</a>
+        <a @click="adminSubmit('3')">Phase 3</a>
+        <a @click="adminSubmit('4')">Phase 4</a>
+        <a @click="adminSubmit('5')">Phase 5</a>
+      </template>
```
pawlh, line 119: What would you think about using something like `<template v-slot:dropdown-items :disabled="adminRepo.length > 0">`? With that we could avoid error handling on lines 90-93 and make the requirement more intuitive

src/main/resources/frontend/src/components/PopUp.vue
```diff
@@ -28,6 +28,7 @@ defineEmits({
   display: flex;
   justify-content: center;
   align-items: center;
+  z-index: 100;
```
pawlh, line 31: :)

src/main/java/edu/byu/cs/server/Server.java
```diff
@@ -67,7 +67,9 @@ public static void setupEndpoints() {
 
                 patch("/user/:netId", userPatch);
 
-                path("/submissions", () -> {
+                post("/submit", adminRepoSubmitPost);
+
+              path("/submissions", () -> {
```
pawlh, line 72: I think the indentation got a little off here

src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -360,7 +360,8 @@ private Submission saveResults(Rubric rubric, int numCommits, int numDaysLate, f
                 score,
                 numCommits,
                 notes,
-                rubric
+                rubric,
+                DaoService.getUserDao().getUser(netId).role() == User.Role.ADMIN
```
pawlh, line 364: Rather than fetching the user from the db twice, I think it'd be better to do it once before the if statement

pawlh (commented):
Took some more time to look through the code and left a few comments.

This is going to be a great feature, thanks for making it happen!

src/main/java/edu/byu/cs/autograder/Grader.java
DallinFromEarth: I'm a fan of making fewer calls to the Database. I'm not sure what the second call you're referring to. In the saveResults() function, there are two calls to the Database, but they're for different tables. One is `DaoService.getQueueDao().get(netId)` and the other is `DaoService.getUserDao().getUser(netId)`. I don't think we call the database for the user any other place in this process, but I could be wrong

pawlh: Yup, you're right, my comment doesn't make much sense. When I looked at this on GitHub I didn't notice that these changes were happening in different functions
image

pawlh: Nevermind 😅

DallinFromEarth: ❤️

src/main/resources/frontend/src/views/AdminView/SubmissionsView.vue
DallinFromEarth: I think this is a brilliant idea! It took a minute to figure it out, but now the dropdown is completely disabled, unless the textbox contains at least `github.com/`

DallinFromEarth:
Changes made! Now the dropdown will be disabled until it contains at least `github.com`. We can talk about how much url verification we want to do on the front end, but I think this is good. Makes it so you can't just put any link in there, but is simple enough.

DallinFromEarth:
Recent merge of main to this branch introduced some changes that need to be fixed with regards to skipping canvas

src/main/java/edu/byu/cs/controller/SubmissionController.java
```diff
@@ -90,7 +91,8 @@ private static void startGrader(String netId, Phase phase, String repoUrl) {
     }
 
     private static void updateRepoFromCanvas(User user, Request req) throws CanvasException {
-        String newRepoUrl = CanvasIntegration.getGitRepo(user.canvasUserId());
+        CanvasIntegration canvas = new CanvasIntegrationImpl();
```
pawlh, line 94: Michael has a getter in `CanvasIntegration`, `getCanvasIntegration()`. If we use that instead of specifying `new CanvasIntegrationImpl()` it will make this code more compatible with mocking

DallinFromEarth: done and done

DallinFromEarth:
> Recent merge of main to this branch introduced some changes that need to be fixed with regards to skipping canvas

These issues are now fixed and admin mode now properly skips over canvas calls

pawlh (approved):

PR #260, opened 2024-03-23
Refactor grader

19mdavenport:
There is very little functionally different in this, I just moved around a bunch of code

PR #261, opened 2024-03-23
Do not decrease scores

19mdavenport:
Base is set to #260 for now as I did it starting from there. 
This changes the decreasing score protection to use rubrics and potential points rather than overall canvas scores.


pawlh (approved):
Seems to work as intended. Thanks for the fix!

pawlh (approved):
Runs as expected 👍🏻

PR #263, opened 2024-03-25
fix score calculation

19mdavenport:
This fixes the displayed grade percentage. Currently, the calculation just drops the percentage by 10*x%, where x is the number of days late. This leads to displaying (not grading, grading works correctly) incorrect scores, like 58% instead of 54% for phase 1 extra credit or 33.97% instead of 41.98% for my branch without any phase 3 unit tests and a dock on code quality

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/score/Scorer.java
+++ b/src/main/java/edu/byu/cs/autograder/score/Scorer.java
@@ -92,7 +92,7 @@ public class Scorer {
 
     private float calculateScoreWithLatePenalty(Rubric rubric, int numDaysLate) throws GradingException {
         float score = getScore(rubric);
-        score -= numDaysLate * PER_DAY_LATE_PENALTY;
+        score *= 1 - (numDaysLate * PER_DAY_LATE_PENALTY);
         if (score < 0) score = 0;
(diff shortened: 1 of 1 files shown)


frozenfrank:
Is this supposed to resolve #168, or is this addressing something different?

frozenfrank:
> This leads to displaying (not grading, grading works correctly) incorrect scores

@19mdavenport 
For those of us who are less familiar with the autograder, can you help me understand the difference between the grading and the scoring of the assignment? Is this suggesting one of the following:
1. The correct grade was being entered into Canvas correctly but displayed incorrectly within the Autograder, OR 
2. The analysis of the results (`Grader`) worked properly, but the results were combined differently than expected to produce the final grade (`Scorer`)
3. Something else

19mdavenport:
@frozenfrank Your first option is correct

19mdavenport:
This was not meant to address any specific github issue. I just noticed a problem and fixed it

pawlh (approved):
> you asked for smaller pull requests? - @19mdavenport 

I asked and you delivered 😎 

Good catch, I didn't notice that these didn't match

PR #265, opened 2024-03-26
fix admin submission to not call canvas

DallinFromEarth:
For some reason #256 had a bug that only appeared on the production server where an admin submission wouldn't complete because it tried calling canvas with the admin's netID. This shouldn't have been happening, and it didn't happen during testing, even after the merges. In theory this PR should fix it, by moving the admin-mode check to the beginning of the Scorer call.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/score/Scorer.java
+++ b/src/main/java/edu/byu/cs/autograder/score/Scorer.java
@@ -40,12 +40,17 @@ public class Scorer {
         rubric = CanvasUtils.decimalScoreToPoints(gradingContext.phase(), rubric);
         rubric = annotateRubric(rubric);
 
+        // skip penalties if running in admin mode
+        if (gradingContext.admin()) {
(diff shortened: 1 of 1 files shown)


frozenfrank (commented):
Interesting. The only lines between where check was and where it is now are the one that calculate the late penalty. 

That code does ask Canvas for the due date of the assignment for the given student. However, I can't explain why it would behave differently in production mode vs the testing environment. Maybe the difference in environment variables causes a different behavior. Or, better, maybe Michael's new Canvas mock is avoiding the API call in testing, but then in production the actual Canvas integration makes the request.

DallinFromEarth:
We have a prediction for what happened:
When running locally, we use the canvas authToken tied to our canvas users, which lets us mess with and view our own due dates. However, in production, everyone uses Paul's canvas authtoken, which we think only lets him get the due date for Paul's canvas user. In production, the feature works for Paul, but not for anyone else.

Regardless, this pr should fix the issue

pawlh (approved):

PR #274, opened 2024-03-28
Use logger instead of System out in `LateDayCalculator`

pawlh:
Currently running the `LateDayCalculator` tests results in lots of warning messages printed to `System.out`. This change allows for us to still log these warnings, but not pollute standard out


frozenfrank (commented):
Looks great! Thanks for fixing this @pawlh !

PR #276, opened 2024-03-28
258 update readme with program arguments (and WSL notes)

### LINKED ISSUE #258: dev: update READ.me with program arguments
A previous PR made it so that the program no longer uses environment variables, but uses program arguments. This needs to be updated on the repo READ.me

ThanGerlek:
Resolves #258.


pawlh (approved):
Looks great, thanks especially for adding Windows specific notes! I bet that many future devs will benefit from those

Issue #230, opened 2024-03-13
Backend: Finish deprecating the numCommits column from the Submissions table


frozenfrank:
This is no longer necessary since it was removed while resolving #160.

While we're at it, also finishing deprecating the `results` field in the Submissions DAO. @pawlh said he's nearly removed it, but we need to verify in the code that all references are cleaned up.

PR #284, opened 2024-03-30
281 backend streamline sql dao column list management

### LINKED ISSUE #281: Backend: Streamline SQL DAO column list management
This process was already begun in commit ec4cd004c08557e08b3ffd41085035e32a19c2a5 of #233 in the Submission DAO. We'll want to finish dynamically generating the proper number of question marks for the insert command.

We'll also want to apply the same techniques to the other DAO's so that they are all on the same page.

frozenfrank:
Resolves #281.


frozenfrank:
@ThanGerlek You asked how this would be done... Behold!

frozenfrank (commented):
It's interesting that the total line count has increased with this PR, even though I was removing code duplication. I attribute this to the extra files and imports I added, as well as the hundreds of lines of JavaDoc comments.

frozenfrank:
This PR has been rebased onto `main` so that these changes can be introduced independent of #233.

At this point, it can be reviewed and merged at will.

frozenfrank:
@DallinFromEarth I'm requesting your review on this issue because you've edited the files recently. 

At one level, you should be aware of these changes for when you will _inevitably_ want to change them again. These changes are designed to simplify and streamline the effort required to change DAO's in the future. With these restructurings, the number of places we'll need to make changes when adding or removing columns, is minimized. Additionally the effort required to create a special case SQL query statement is dramatically reduced. Does this actually help, or are there additional changes you would like to see related to this topic?

At another level, I want you to review my code. Have I properly transformed the existing code into this new format? Are there other changes or adjustments you would like to see before we merge it?

I welcome your comments :)

frozenfrank:
@pawlh This passes all the checks :) 

While reviewing, notice that the commit history has been intentionally authored with distinct changes each in their own commit. You may appreciate viewing the changes commit by commit as they each describe why the changes were made.

src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
```diff
@@ -12,69 +14,63 @@
...
-            statement.setString(1, submission.netId());
-            statement.setString(2, submission.repoUrl());
-            statement.setTimestamp(3, Timestamp.from(submission.timestamp()));
-            statement.setString(4, submission.phase().toString());
-            statement.setBoolean(5, submission.passed());
-            statement.setFloat(6, submission.score());
-            statement.setString(7, submission.headHash());
-            statement.setInt(8, submission.numCommits());
-            statement.setString(9, submission.notes());
-            statement.setString(10, "{}");
```
frozenfrank, line 36: Note that this removes some references of `Submission.Results` though it doesn't fully deprecate the field. See #230

ThanGerlek:
Well, it's exactly what I was thinking of when I said it'll be more complex than necessary 😂, but it looks really good! Guess I was being too much of a pessimist--nice job!

src/main/java/edu/byu/cs/dataAccess/sql/helpers/SqlReader.java
```diff
@@ -136,14 +136,15 @@ private void setValue(PreparedStatement ps, int wildcardIndex, T item, ColumnDef
...
         // This represents all the supported column types
-        if (value == null) ps.setNull(wildcardIndex, NULL);
-        else if (value instanceof String v) ps.setString(wildcardIndex, v);
-        else if (value instanceof Integer v) ps.setInt(wildcardIndex, v);
-        else if (value instanceof Float v) ps.setFloat(wildcardIndex, v);
-        else if (value instanceof Boolean v) ps.setBoolean(wildcardIndex, v);
-        else if (value instanceof Timestamp v) ps.setTimestamp(wildcardIndex, v);
-        else if (value instanceof Object v) ps.setObject(wildcardIndex, v);
-        else throw new RuntimeException("Unsupported type of value: " + value);
+        switch (value) {
```
frozenfrank, line 139: @pawlh I figured out that my environment didn't suggest this change because it was running Java 20. In fact, it complained about this feature because this language feature was introduced in Java 21. After I upgraded my version to Coretto 21, it worked just great.

As long as we're clear that Java 21 is the oldest version we support, then this is definitely a much cleaner syntax than the previous edition.

pawlh (approved):
Aside from my one note about annotations, I think everything looks great, so here's an approval ✅ 

I think this is probably way more complicated/sophisticated a solution than a project of this scale needs and it might slightly hurt future maintainability (when it comes to devs getting started with the project), but in the name of creating cool things I think that's okay

src/main/java/edu/byu/cs/dataAccess/sql/helpers/SqlReader.java
```diff
@@ -0,0 +1,341 @@
+package edu.byu.cs.dataAccess.sql.helpers;
+
+import edu.byu.cs.dataAccess.DataAccessException;
+import edu.byu.cs.dataAccess.sql.SqlDb;
+import org.eclipse.jgit.annotations.NonNull;
+import org.eclipse.jgit.annotations.Nullable;
```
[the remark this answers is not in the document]
frozenfrank: After looking into this issue further, it seems that finding the right import is actually quite difficult. It seems there are plenty of people looking for an answer, but no one providing a standardized and maintained project for this.

See this [StackOverflow question](https://stackoverflow.com/questions/4963300/which-notnull-java-annotation-should-i-use) with 1300+ upvotes, a detailed synopsis of the lay of the land, but no satisfactory conclusion.

Based on the discussion in the thread above, it's clear that no standard answer exists. It seems the answer is mostly a matter of personal preference, and integration with whichever tools a particular project is depending on. Also, its seems pretty straightforward to add a maven dependency and search-and-replace a few `import` statements if we ever want to switch.

Since this project has no precedent, I'm going to assume that this approach is fine for now.

frozenfrank:
> I think this is probably way more complicated/sophisticated a solution than a project of this scale needs and it might slightly hurt future maintainability (when it comes to devs getting started with the project), but in the name of creating cool things I think that's okay

@pawlh Thanks for the approval, and thanks for sharing your concerns about over-complexity. I've attempted to address your concerns below.

## Future Maintainability
You bring up a good point that future developers will not be familiar with this system unique to our project. Hopefully, they'll be able to see that the helper methods are simply wrappers over the boilerplate JDBC code they would be familiar with writing, and recognize that writing out the full SQL query with JDBC code is still valid. I've already done my best to carefully document each of the methods and their call signatures with all the details a future developer would need to figure it out.

To further support future developers, I've also added an additional `dataAccess/README.md` file that provides an introduction to this unique and innovative system.

## Over Complicated
The `SqlReader` class, which is really just a set of helper methods, was created not to complicate SQL code, but to simplify it. Rather than being built merely to be a "cool thing," it was introduced  for the following reasons:
1. To centralize the definition of SQL column definitions, therefore reducing the possibility of accidentally not updating some occurrences of the `SELECT` statements.
2. Simplifying the developer experience when writing new queries. Following the best practices of abstracting away common, repeated parts of software development, the "try-with-resources" blocks and the error handling and the actual preparing of `PreparedStatements` has been centralized, and each method is required only to define the actual logic that is unique to the method.

PR #287, opened 2024-04-02
edit get message time

19mdavenport:
Do you think this is a good idea to add? This would hopefully prevent students from timing out on the phase 6 tests and not needing to let tests run for 3.5 minutes.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -6,8 +6,14 @@ import edu.byu.cs.util.FileUtils;
 import edu.byu.cs.util.ProcessUtils;
 
 import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 
 public class CompileHelper {
+
+    private static final int WEBSOCKET_MESSAGE_TIME = 250;
+
     private final GradingContext gradingContext;
 
     public CompileHelper(GradingContext gradingContext) {this.gradingContext = gradingContext;}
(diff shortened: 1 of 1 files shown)


src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
```diff
@@ -92,4 +99,18 @@ private String getMavenError(String output) {
         }
         return builder.toString();
     }
+
+    private void replaceGetMessageTime() throws GradingException {
+        String regex = "return [0-9]*L;";
```
pawlh, line 104: Shouldn't this be`[0-9]+L`?

src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
```diff
@@ -92,4 +99,18 @@ private String getMavenError(String output) {
...
+        String replace = "return " + WEBSOCKET_MESSAGE_TIME + "L;";
+        Path path = new File(gradingContext.stageRepo(),
+                "/shared/src/test/java/passoffTests/TestFactory.java").toPath();
+        try {
+            String orig = Files.readString(path, StandardCharsets.UTF_8);
+            String replaced = orig.replaceAll(regex, replace);
+            Files.writeString(path, replaced);
+        } catch (Exception e) {
+            throw new GradingException("Failed to modify TestFactory", e);
```
pawlh, line 113: Are students allowed to move this file? If not that makes things easier, but either way I think we should give a better error message, being more explicit about what is missing from where

pawlh (commented):
I left a couple couple, but I think this is a good change

src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
19mdavenport: Hmm. I wouldn't think so, but I'm not sure. I guess there was a way easier way to change this too.

PR #288, opened 2024-04-02
Edit wait time

19mdavenport:
Way easier way of doing what the previous pull request did


pawlh:
Definitely wayyy easier, but the downside is that we will no longer be storing tests identical to the starter code. That means that if there are changes to `WebSocketTests.java` in the future, we'll need to remember to change this line each time when updating the autograder.

What do you think?

pawlh (approved):

PR #289, opened 2024-04-02
Streamline phase configuration

frozenfrank:
## What is going on here?
Reorganizes the phase configuration data to leverage Java's object oriented model. Instead of using switch statements to provide information outside of the `Phase` object, that data is now being stored _in_ the `Phase` object.

So far, the existing getters are simply being changed internally to read the data differently. By the time this PR is finished, the entire `PhaseUtils` class will be ripped out and the code will be calling methods on the `Phase` object instead.

## Why go through the effort?
This shift, when complete, will make simplify the lives of future maintainers by having all related fields in one place which reduces the likelihood of a future developer forgetting to change some values.

Additionally, when stored as objects, it becomes a lot easier to make a leap and just store all of the data in a table instead of hardcoding configuration values in the source code files. Once the data is in a table, it also opens the door for settings pages and pretty GUI's to interact with the values and enable professors and more kinds of authorized people access to these controls.

Therefore, I see this as the first step in a longer process to have a robust, dynamic system on our hands that will outlive all of us as TA's.

## Todo List
- [ ] Move value getting methods directly into `Phase`
- [ ] Convert `Phase` into an `enum` which will enable dynamically loading
- [ ] Update all usages of `PhaseUtils` to use methods of `Phase`
- [ ] Make fields `private` instead of `public` after the methods exist for reading them
- [ ] Write JavaDoc documentation for each of the fields describing their purpose and nuances
- [ ] Eliminate `PhaseUtils`
- [ ] Eliminate nearly identical fields `Phase.stringValue` and `Phase.stringName`. They both exist because of a funky configuration in the `Quality` phase (it has `stringValue: "Quality"`, and `stringValue: "42"`). If we can reconcile to make them the same, then the rest of the duplication wouldn't be necessary. I didn't understand enough about why that separation existed in order to adjust it myself.
- [ ] Create a separate issue for storing and decoding all the data dynamically from a table (see #156).

### DIFF
--- a/src/main/java/edu/byu/cs/model/Phase.java
+++ b/src/main/java/edu/byu/cs/model/Phase.java
@@ -3,6 +3,8 @@ package edu.byu.cs.model;
 import java.util.Map;
 import java.util.Set;
 
+import static edu.byu.cs.model.Rubric.RubricType.*;
+
 public enum Phase {
 
     // FIXME: Load this list of data dynamically from a configuration table #156
@@ -13,7 +15,7 @@ public enum Phase {
             true,
             null,
             880445,
-            Map.of(Rubric.RubricType.PASSOFF_TESTS, "_1958"),
+            Map.of(PASSOFF_TESTS, "_1958"),
             -1,
             "shared",
             null,
@@ -26,7 +28,7 @@ public enum Phase {
             true,
             Phase0,
             880446,
-            Map.of(Rubric.RubricType.PASSOFF_TESTS, "_1958"),
+            Map.of(PASSOFF_TESTS, "_1958"),
             -1,
             "shared",
             null,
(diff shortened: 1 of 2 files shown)


frozenfrank:
@19mdavenport @pawlh I'm just starting down this path, but I recognize that this is a radical change in architecture of the current project that will end up touching lots of files. 

I wanted to show you my progress so far and see what you think. Is this a course I should continue pursuing? Alternatively, if you don't like these changes, I would just stop now.

19mdavenport:
I like this idea in general

frozenfrank:
> I like this idea in general

@19mdavenport And are there some specifics about it that you want to comment on so far?

19mdavenport:
Closing stale pull request. This can be reopened later if we decide this is still worthwhile

PR #297, opened 2024-04-04
Add process utils tests

pawlh:
# What
This PR adds unit tests for `ProcessUtils`. Super exciting 😉


pawlh:
@19mdavenport `ProcessUtils` is your baby, so let me know if you think this has sufficient coverage

pawlh:
Oof, 20.3 seconds for a single set of tests. Unfortunately I'm not sure how to avoid it with this one, as the size of it is the whole point of the test. Maybe the size of the input could be reduced while still being large enough to test the buffer draining?

image

19mdavenport (approved):
Seems like pretty good coverage to me

PR #308, opened 2024-04-11
Index submission table

19mdavenport:
And two other things I think will be necessary or nice


frozenfrank (commented):
Changes look good to me. I'm assuming that the table indexing will improve read performance for certain queries.

I haven't run the code myself to give an explicit approval though.

pawlh (approved):

PR #327, opened 2024-04-24
filter with index before sorting

pawlh:
# What this does
This PR revises the query that fetches the latest submission for each student-phase. Arguably we don't need this feature anymore, but given that it is still present I wanted to patch our memory overflow error:

`ERROR 1038 (HY001): Out of sort memory, consider increasing server sort buffer size`

I tried <ins>a lot</ins> of different ways to get the query to fallback on indexes, but the conclusion of my research was that indexes are hard.

This PR modifies the query to filter out all the submissions we don't care about _before_ sorting. This technically doesn't fix the problem and mostly just moves the issue further down, but I'd be pretty surprised if we encounter it again, especially if intend on clearing the database each semester

### DIFF
--- a/src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
+++ b/src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
@@ -9,14 +9,11 @@ import edu.byu.cs.dataAccess.SubmissionDao;
 import edu.byu.cs.model.Phase;
 import edu.byu.cs.model.Submission;
 
-import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.time.Instant;
 import java.util.*;
-
-import static java.sql.Types.NULL;
 
 public class SubmissionSqlDao implements SubmissionDao {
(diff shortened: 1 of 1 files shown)


src/main/java/edu/byu/cs/dataAccess/sql/SubmissionSqlDao.java
```diff
@@ -80,30 +79,43 @@ public Collection<Submission> getAllLatestSubmissions() throws DataAccessExcepti
...
+                    """
+                            SELECT s.net_id, s.repo_url, s.timestamp, s.phase, s.passed, s.score, s.num_commits, s.head_hash, s.notes, s.rubric, s.admin
+                            FROM submission s
+                            INNER JOIN (
+                                SELECT net_id, phase, MAX(timestamp) AS max_timestamp
+                                FROM submission
+                                GROUP BY net_id, phase
+                            ) s2 ON s.net_id = s2.net_id AND s.phase = s2.phase AND s.timestamp = s2.max_timestamp
+                            ORDER BY s2.max_timestamp DESC
+                            """ +
```
pawlh, line 93: @frozenfrank 
I didn't foresee this issue when reviewing your big SQL PR, but I realized with this that your select helper doesn't cover queries that use `join`

Issue #300, opened 2024-04-04
fullstack: refactor RubicItems to be dynamic


DallinFromEarth:
Currently a `Rubric` object contains 3 `RubricItem` objects. To be more flexible, we should refactor this to be some sort of list of `RubricItem`s or other system where it is easy to add and remove RubricItems

PR #328, opened 2024-04-24
Dev skip project compilation

frozenfrank:
This minor adjustment gives DEV's the ability to skip compilation of projects as needed. In the process, I also broke the existing code into smaller functions which allows us to skip part of their functionality.

I'm currently using this to test #233 since I'm not interested in the actual grading results, and future dev's may want this option as well.

Ideally, there would exist a bank of options that allow developers the ability to skip any combination of the project evaluation based on their particular project. Ideally, those options would be available in some config object or ENV variable or command line option. However, since we don't have a whole system set up yet, this is only a lightweight adjustment to provide this functionality via a hard-coded local variable.


src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -112,6 +89,45 @@ public void run() {
         }
     }
 
+    private Map<String, Rubric.Results> evaluateProject(RubricConfig rubricConfig) throws GradingException {
+        // NOTE: Ideally these would be treated with enum types. That will need to be improved with #300.
```
frozenfrank, line 93: Note that there is still lots of room for improvement as in #300, but this is only a lightweight adjustment for now.

pawlh (approved):
I'm cool with this as a temporary approach once the tests are passing again

PR #329, opened 2024-04-25
spring quality updates

19mdavenport:
Updates to the quality checker for spring 2024

Adds checks for package names and constant names
Removes test case exclusion
Simplifies grading so each category is all or nothing

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
+++ b/src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
@@ -1,7 +1,6 @@
-package edu.byu.cs.autograder.quality;
+package edu.byu.cs.autograder;
 
 import com.google.gson.Gson;
-import edu.byu.cs.autograder.GradingException;
 import edu.byu.cs.util.ProcessUtils;
 
 import java.io.File;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
```diff
@@ -30,36 +29,47 @@ public class QualityAnalyzer {
...
-    public QualityOutput runQualityChecks(File stageRepo) throws GradingException {
+    /**
+     * Runs quality checks for a repository
+     *
+     * @param stageRepo file to the repo to check
+     * @return QualityAnalysis object containing score, results, and notes
+     */
+    public QualityAnalysis runQualityChecks(File stageRepo) throws GradingException {
         ProcessBuilder processBuilder = new ProcessBuilder().directory(stageRepo.getParentFile())
                 .command("java", "-jar", checkStyleJarPath, "-c", "cs240_checks.xml", "repo");
```
pawlh, line 40: where is `cs240_checks.xml` located?

pawlh (approved):

src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
19mdavenport: It's inside the jar

PR #331, opened 2024-04-25
remove percentage when submission failed

### LINKED ISSUE #299: frontend: hide score from student if the submission fails
Currently, if a submission fails, it doesn't get sent to canvas. BUT it still shows a grade to the student, which can be confusing, leading them to think they got given a score when they haven't. 

### Proposal
If `!submission.passed` then don't show the score, just say failed

19mdavenport:
This removes the percentage display from the past submissions if the submission failed.

I believe this resolves #299 unless there is a different area that is still displaying scores, but I think we've been hiding that for a while.


frozenfrank (approved):
## Overall
Thanks for making this adjustment.

I did add a commit to the branch that removes a redundant method call. The `.toString()` is not necessary because [Number.toPrecision()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Number/toPrecision) already returns a string result.

## Thoroughness Search
My search for other occurrences turned up dry. None of the following regexes returned any relevant results:
* `(?<!submission)\.score`
    * Returns only results in the notes/rubrics sections of the code.
* `submission\.score`
    * Returned results are shown only in the admin section, or are already specifically guarded to show only under passing conditions.

## Therefore
This looks good to me!

PR #332, opened 2024-04-25
Remove old logs code

19mdavenport:
This removes code used for the previous logging system that is now unused.


.gitignore
```diff
@@ -38,5 +38,4 @@ build/
 ### temp files ###
 *.class
 /tmp-*
-logs/
```
pawlh, line 41: let's actually keep excluding this. I have `logback` set to still dump some items into a log file for redundancy

https://github.com/softwareconstruction240/autograder/blob/339f290fb39f0afd92d7e621c80c510fb8841d35/src/main/resources/logback.xml#L22

pawlh (changes requested):

pawlh (approved):

PR #334, opened 2024-04-25
Remove phase 42

19mdavenport:
This removes phase 42 for quality and keeps track of phases using enums instead of strings and ints


src/main/java/edu/byu/cs/controller/SubmissionController.java
pawlh, line 1: Line 203 should be a try-catch that uses `Phase.valueOf(phase)` instead of `PhaseUtils.getPhaseByString(phase)`. Right now this endpoint is generating a 400 error

pawlh: maybe
```
Phase phaseEnum = null;
try {
    phaseEnum = Phase.valueOf(phase);
} catch (IllegalArgumentException e) {
    LOGGER.error("Invalid phase", e);
    halt(400, "Invalid phase");
}
```

pawlh (changes requested):

src/main/java/edu/byu/cs/controller/SubmissionController.java
19mdavenport: I didn't think about that endpoint. Thank you

pawlh (approved):

PR #347, opened 2024-05-02
125 accept divserse repo urls

### LINKED ISSUE #125: backend: Accept a diversity of GitHub URLs
## Url Submission Examples
[Category 1] Currently, the only accepted forms of GitHub URLS are the following
* `https://github.com/<USERNAME>/<REPO_NAME>`
* `https://github.com/<USERNAME>/<REPO_NAME>.git`

[Category 2] However, this is also a common variant that students end up submitting for their REPO:
* `https://github.com/<USERNAME>/<REPO_NAME>/tree/main`
* `https://github.com/<USERNAME>/<REPO_NAME>?tab=readme-ov-file`

[Category 3] Or, they incorrectly post a link to a folder inside of the GitHub repository, like the following:
* `https://github.com/<USERNAME>/<REPO_NAME>/tree/main/0-chess-moves/starter-code/chess` (specific branch)
* `https://github.com/<USERNAME>/<REPO_NAME>/pull/1` (pull request instead)

## Issue
The urls in Category 2 really should be just as valid as those in category 1. They are still representing the same information, and a student is likely to end up grabbing this url. However, the autograder will fail.

## Proposal
Adapt the autograder to extract the url _only up to_ the name of the repository, and ignore all the trailing characters. This would simplify issues that really shouldn't exist for the students. 

It would also come with the side effect that students would be **quietly** disallowed from pointing the autograder at a specific branch (which they may attempt to do intentionally). Here are several solutions to this particular issue:
1. Display the _effective URL_ inside the autograder. A sharp student will be able to notice the difference and merge their code into `main`.
2. Specifically display a message when we use a normalized URL instead of the provided URL. Since they'll be using the same url all semester, and the point is to remove the need to update their submission inside Canvas, this may not be the recommended approach.

## Possible Implementation Note
This kind of cleanup could be performed while setting the `repoURL` param on a `Grader` object.

Fiwafoofa:
Strips off trailing characters in github URLs for non-admin submissions. 
Supports SSH urls. 
Addeds test for `cleanRepoUrl` method.

Resolves #125.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/Grader.java
+++ b/src/main/java/edu/byu/cs/autograder/Grader.java
@@ -50,7 +50,7 @@ public class Grader implements Runnable {
      * @param observer the observer to notify of updates
      * @param phase    the phase to grade
      */
-    public Grader(String repoUrl, String netId, Observer observer, Phase phase, boolean admin) throws IOException {
+    public Grader(String repoUrl, String netId, Observer observer, Phase phase, boolean admin) throws IOException, GradingException {
         // Init files
         if (!admin) {
             repoUrl = cleanRepoUrl(repoUrl);
@@ -159,9 +159,9 @@ public class Grader implements Runnable {
      *
      * @param repoUrl The student's repository URL.
      * @return Cleaned URL with everything after the repo name stripped off.
-     * @throws IOException Throws IOException if repoUrl does not follow expected format
+     * @throws GradingException Throws IOException if repoUrl does not follow expected format
      */
-    public static String cleanRepoUrl(String repoUrl) throws IOException {
+    public static String cleanRepoUrl(String repoUrl) throws GradingException {
         String[] regexPatterns = {
             "https?://github\\.com/([^/?]+)/([^/?]+)", // https
             "git@github.com:([^/]+)/([^/]+).git" // ssh
@@ -179,7 +179,7 @@ public class Grader implements Runnable {
                 return String.format("https://github.com/%s/%s", githubUsername, repositoryName);
             }
         }
-        throw new IOException("Could not find github username or repository name given '" + repoUrl + "'.");
+        throw new GradingException("Could not find github username and repository name given '" + repoUrl + "'.");
     }
 
     public interface Observer {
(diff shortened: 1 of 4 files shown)


src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -148,6 +153,34 @@ private Rubric assembleResultsToRubric(RubricConfig rubricConfig, Map<String, Ru
...
+     * unless it ends in `.git`.
+     *
+     * @param repoUrl The student's repository URL.
+     * @return Cleaned URL with everything after the repo name stripped off.
+     * @throws GradingException Throws IOException if repoUrl does not follow expected format
+     */
+    public static String cleanRepoUrl(String repoUrl) throws GradingException {
+        String[] regexPatterns = {
+            "https?://github\\.com/([^/?]+)/([^/?]+)", // https
+            "git@github.com:([^/]+)/([^/]+).git" // ssh
```
frozenfrank, line 167: Nice work handling SSH urls as well!

src/test/java/edu/byu/cs/autograder/GraderTest.java
frozenfrank, line 1: Thanks for writing tests!

I'm glad the issues was written well enough to use for testing purposes 😆

src/test/java/edu/byu/cs/autograder/GraderTest.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Test
+    @Tag("cleanRepoUrl")
+    @DisplayName("Admin submissions are not cleaned")
+    void adminSubmissionsAreNotCleaned() throws GradingException, IOException {
+        String originalUrl = "https://github.com/<USERNAME>/<REPO_NAME>/tree/main/0-chess-moves/starter-code/chess";
+        // This cannot be easily tested since `Grader` has many other dependencies that would need to be organized as well.
+        // As this point, we are not spending the time to make `Grader` into a more testable format.
+//        var grader = new Grader(originalUrl, "student_id", null, null , true);
```
frozenfrank, line 73: Rip that this doesn't work.

frozenfrank (approved):
Great work, @Fiwafoofa !

PR #348, opened 2024-05-02
Disable admin commit verification

19mdavenport:
Previously, the git verification ran for admins, but just was never reported when scoring. However, the commit verification could throw exceptions if the admin had submitted with a different repository in the past. This completely disables the verification for admins so the exception isn't thrown.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -44,8 +44,8 @@ public class GitHelper {
         File stageRepo = gradingContext.stageRepo();
         fetchRepo(stageRepo);
 
-        boolean gradedPhase = PhaseUtils.isPhaseGraded(gradingContext.phase());
-        return (gradedPhase  && !gradingContext.admin()) ?
+        boolean requiresVerification = PhaseUtils.isPhaseGraded(gradingContext.phase()) && !gradingContext.admin();
+        return requiresVerification ?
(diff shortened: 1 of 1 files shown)


src/main/java/edu/byu/cs/autograder/git/GitHelper.java
```diff
@@ -142,7 +143,7 @@ private CommitVerificationResult verifyRegularCommits(Git git, Collection<Submis
 
         CommitsByDay commitHistory = CommitAnalytics.countCommitsByDay(git, lowerThreshold, upperThreshold);
         CommitVerificationResult commitVerificationResult = commitsPassRequirements(commitHistory);
-        LOGGER.debug("Commit verification result: " + JSON.toString(commitVerificationResult));
+        LOGGER.debug("Commit verification result: {}", JSON.toString(commitVerificationResult));
```
frozenfrank, line 146: Thanks for fixing this printing issue. What was wrong with the way I had it?

frozenfrank (approved):
Looks great. This is a great place to make this modification.

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
19mdavenport: Supposedly this change might make it faster? I think java string concatenation runs in linear time rather than constant time and this may or may not make it marginally faster. (Mostly I wanted to remove an IntelliJ warning)

PR #354, opened 2024-05-03
UI Fixes

19mdavenport:
This contains some fixes and improvements for the UI


frozenfrank (commented):
These look good to me. I like the use of icons ❗

PR #357, opened 2024-05-03
Fix commit verification null exception

frozenfrank:
## Overview
This fixes the null pointer exception that occurred in the commit verification process when students resubmitted their phase after a blocked submission but before it was manually approved.

### Complete Solution
Beyond simply fixing the error, it also introduces improved error handling so that future errors will be caught and reported, but grading will still continue and the submission can still be saved in the database. Additionally, the call signatures of related functions have been adjusted to minimize the possibility of null pointer issues in the future.

Key adjustments are the titles of each commit in the PR.

## Notifying Error Message
This is the sample exception it will resolve:
```bash
java.lang.NullPointerException: Cannot invoke "edu.byu.cs.model.Submission$ScoreVerification.penaltyPct()" because "scoreVerification" is null
        at edu.byu.cs.autograder.git.GitHelper.preserveOriginalVerification(GitHelper.java:128)
        at edu.byu.cs.autograder.git.GitHelper.verifyCommitRequirements(GitHelper.java:88)
        at edu.byu.cs.autograder.git.GitHelper.setUp(GitHelper.java:49)
        at edu.byu.cs.autograder.Grader.run(Grader.java:86)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -34,19 +34,41 @@ import java.util.Collection;
 public class GitHelper {
     private static final Logger LOGGER = LoggerFactory.getLogger(GitHelper.class);
     private final GradingContext gradingContext;
+    private String headHash;
 
     public GitHelper(GradingContext gradingContext) {
         this.gradingContext = gradingContext;
     }
 
     // ## Entry Point ##
-    public CommitVerificationResult setUp() throws GradingException {
+    public CommitVerificationResult setUpAndVerifyHistory() throws GradingException {
+        setUp();
+        return verifyCommitHistory();
+    }
+    public void setUp() throws GradingException {
         File stageRepo = gradingContext.stageRepo();
-        fetchRepo(stageRepo);
+        fetchRepo(gradingContext.stageRepo());
+        headHash = getHeadHash(stageRepo);
+    }
+    public CommitVerificationResult verifyCommitHistory() {
+        if (headHash == null) {
+            throw new RuntimeException("Cannot verifyCommitHistory before headHash has been populated. Call setUp() first.");
+        }
 
-        boolean requiresVerification = PhaseUtils.isPhaseGraded(gradingContext.phase()) && !gradingContext.admin();
-        return requiresVerification ?
-                verifyCommitRequirements(stageRepo) : skipCommitVerification(stageRepo);
+        File stageRepo = gradingContext.stageRepo();
+        try {
+            boolean requiresVerification = PhaseUtils.isPhaseGraded(gradingContext.phase()) && !gradingContext.admin();
+            if (!requiresVerification) {
+                return skipCommitVerification(true, headHash, null);
+            }
+            return verifyCommitRequirements(stageRepo);
+        } catch (GradingException e) {
+            // Grading can continue, we'll just alert them of the error.
+            String errorStr = "Internally failed to evaluate commit history: " + e.getMessage();
+            gradingContext.observer().update(errorStr);
+            LOGGER.error("Failed to evaluate commit history", e);
+            return skipCommitVerification(false, headHash, errorStr);
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/java/edu/byu/cs/autograder/git/GitHelper.java
```diff
@@ -34,19 +34,47 @@
...
+    public void setUp() throws GradingException {
         File stageRepo = gradingContext.stageRepo();
-        fetchRepo(stageRepo);
+        fetchRepo(gradingContext.stageRepo());
+        headHash = getHeadHash(stageRepo);
+    }
+    public CommitVerificationResult verifyCommitHistory() {
+        if (headHash == null) {
+            throw new RuntimeException("Cannot verifyCommitHistory before headHash has been populated. Call setUp() first.");
+        }
```
19mdavenport, line 56: Is it possible headHash could be null for any other reason? Would you be ok making this a GradingException instead of a RuntimeException?

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
```diff
@@ -71,12 +99,18 @@ private void fetchRepo(File intoDirectory) throws GradingException {
...
-    private CommitVerificationResult skipCommitVerification(File stageRepo) throws GradingException {
-        LOGGER.debug("Skipping commit verification");
+    private CommitVerificationResult skipCommitVerification(boolean verified, File stageRepo) throws GradingException {
         String headHash = getHeadHash(stageRepo);
+        return skipCommitVerification(verified, headHash, null);
+    }
+    private CommitVerificationResult skipCommitVerification(boolean verified, @NonNull String headHash, String failureMessage) {
+        if (headHash == null) {
+            throw new IllegalArgumentException("Head hash cannot be null");
+        }
```
19mdavenport, line 109: Same question about headHash being null. I'd rather not have a student end up with an error message that just says "Head hash cannot be null" and wondering what happened

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
frozenfrank: ## Head Hash nullability
That's a good question about the head hash. 

Head hash is already defined as a `NOT NULL` field in the SQL database, and is required in many areas of the application.

In practice, the head has really can't ever be `null`. A git repo always has a head. This check is mostly verifying that the `setUp()` method has been called before attempting to run the method. I'm providing the developer with a prettier error message if they forget to call the set up the object first. Currently, the `setUp()` method is not ever called from outside the class because the `Grader` now calls `CommitVerificationResult commitVerificationResult = gitHelper.setUpAndVerifyHistory();`

See the create table statement for reference.
```sql
 createSubmissionTableStatement.executeUpdate("""
                        CREATE TABLE IF NOT EXISTS `submission` (
                                `id` INT NOT NULL AUTO_INCREMENT,
                                `net_id` VARCHAR(20) NOT NULL,
                                `repo_url` VARCHAR(200) NOT NULL,
                                `head_hash` VARCHAR(40) NOT NULL,
                                `timestamp` DATETIME NOT NULL,
                                `phase` VARCHAR(9) NOT NULL,
                                `passed` BOOL NOT NULL,
                                `score` FLOAT NOT NULL,
                                `notes` TEXT,
                                `rubric` JSON,
                                `verified_status` VARCHAR(30),
                                `verification` JSON,
                                `admin` BOOL NOT NULL,
                                PRIMARY KEY (`id`),
                                INDEX sort_index (timestamp),
                                CONSTRAINT `net_id`
                                    FOREIGN KEY (`net_id`)
                                    REFERENCES `user` (`net_id`)
                                    ON DELETE CASCADE
                                    ON UPDATE CASCADE
                            ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci
                        """);
```

## Throwing a Runtime Exception
One of the points of this newly sub-divided method was to not throw any checked exception. I wanted this to catch all errors inside it, notify the observer, and then continue with the grading. This allows a submission to proceed and be logged in the database as opposed to throwing an error and halting early like happened earlier today.

I'm not sure how a student could end up with a repo without a head hash. In this case, the error isn't checking for student setup issues, but this is meant to report to a developer if create the preconditions properly. A checked exception isn't the right kind of exception to throw in that case, I want this to fail fast during development if a developer does something wrong. In this particular case, it's never something that a developer would be able to recover from.

In the imaginary situation where a student does end up without a head hash, that would be a fatal error anyways. We can't checkout code, grade anything without it. A `RuntimeException` that will halt the grading is appropriate.

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
frozenfrank: If I don't but the check here, the SQL will throw the for me when it attempts to save in the database. This check just allows the computer to print out a helpful stack trace with a named error to help a developer track down the actual error.

19mdavenport (approved):

PR #360, opened 2024-05-08
check for common issues and provide helpful error messages - package structure and module interdependence

### LINKED ISSUE #365: Backend: Identify Common Issues — Bad Repo Structure
This is a sub-issue of #266 that describes and defines one issue that can be solved independent of the others.

## Overview
* Affects phases: all
* Root cause: Students didn't set up their repository correctly.
* Prevents error: "Failed to Package Repo"
* Tip on detection: 
     > Your code should have all of the following modules. You may have placed this in the wrong location, or set the wrong folder as the repository root. You are missing the following: ...

## Detection Strategy
Define a list of folders that must exist in the student's codebase. Verify that they all exist, and if not, provide a helpful error message. This strategy is valid since we provide most of the folder structure out of the gate in the starter code.

## Implementation Details

The files should have **all** the following: 
* `server/src/main/java`
* `server/src/main/resources`
* `server/src/test/java`
* `client/src/main/java`
* `client/src/test/java`
* `shared/src/main/java`
* `shared/src/test/java`

### LINKED ISSUE #366: Backend: Identify Common Issues — Client/Server Interdependency
This is a sub-issue of #266 that describes and defines one issue that can be solved independent of the others.

## Overview
* Affects phases: 4
* Root Cause: Students attempted to use classes or objects in their server from the client, or vice versa.
* Prevents error: "Failed to Package Repo"
* Tip on detection: 
     > The `client` and `server` can both import from `shared` module, but they cannot import from each other. Make sure any resources needed by _both_ the `client` and `server` and located in the `shared` module.

The fail to compile error already contains the exact output, and if students looked carefully, they would be able to figure this one out. However, they often don't realize that they crossed significant boundaries. Students who run into this error usually fail to understand how separate and distinct the `client` and `server` are from each other; they think of them as just one big project, which it's not.

## Detection Strategy
Scan the packaging output for specific files importing objects from the `client` or `server` module inappropriately

## Implementation Details
When there is a problem packaging the repo, scan the error output for regexes that indicate this issue. Then warn the student that this is an issue they need to fix. We can still refer to the generated error output for the actual line and column definitions of the packages not allowed.

19mdavenport:
Draft PR so people know what I'm working on and collaborate if needed

resolves #365
resolves #366

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -4,6 +4,7 @@ import edu.byu.cs.autograder.GradingContext;
 import edu.byu.cs.autograder.GradingException;
 import edu.byu.cs.util.ProcessUtils;
 
+import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
 
@@ -15,15 +16,20 @@ public class CompileHelper {
     }
 
     private final Collection<StudentCodeInteractor> currentVerifiers =
-            List.of(new ProjectStructureVerifier());
+            List.of(new ProjectStructureVerifier(), new ModuleIndependenceVerifier());
 
     private final Collection<StudentCodeInteractor> currentModifiers =
             List.of(new PomModifier(), new PassoffJarModifier(), new TestFactoryModifier());
 
     public void compile() throws GradingException {
-        for(StudentCodeInteractor verifier : currentVerifiers) verifier.interact(gradingContext);
-        for(StudentCodeInteractor modifier : currentModifiers) modifier.interact(gradingContext);
-        packageRepo();
+        try {
+            StudentCodeReader reader = StudentCodeReader.from(gradingContext);
+            for(StudentCodeInteractor verifier : currentVerifiers) verifier.interact(gradingContext, reader);
+            for(StudentCodeInteractor modifier : currentModifiers) modifier.interact(gradingContext, reader);
+            packageRepo();
+        } catch (IOException e) {
+            throw new GradingException("Failed to read project contents", e);
+        }
     }
 
 
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -15,23 +15,34 @@ public class CompileHelper {
         this.gradingContext = gradingContext;
     }
 
-    private final Collection<StudentCodeInteractor> currentVerifiers =
+    private final Collection<StudentCodeVerifier> currentVerifiers =
             List.of(new ProjectStructureVerifier(), new ModuleIndependenceVerifier());
 
-    private final Collection<StudentCodeInteractor> currentModifiers =
+    private final Collection<StudentCodeModifier> currentModifiers =
             List.of(new PomModifier(), new PassoffJarModifier(), new TestFactoryModifier());
 
     public void compile() throws GradingException {
+        verify();
+        modify();
+        packageRepo();
+    }
+
+    public void verify() throws GradingException {
         try {
             gradingContext.observer().update("Verifying code...");
 
             StudentCodeReader reader = StudentCodeReader.from(gradingContext);
-            for(StudentCodeInteractor verifier : currentVerifiers) verifier.interact(gradingContext, reader);
-            for(StudentCodeInteractor modifier : currentModifiers) modifier.interact(gradingContext, reader);
-
-            packageRepo();
+            for(StudentCodeVerifier verifier : currentVerifiers) {
+                verifier.verify(gradingContext, reader);
+            }
         } catch (IOException e) {
             throw new GradingException("Failed to read project contents", e);
+        }
+    }
+
+    public void modify() throws GradingException {
+        for(StudentCodeModifier modifier : currentModifiers) {
+            modifier.modify(gradingContext);
         }
     }
 
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -23,9 +23,12 @@ public class CompileHelper {
 
     public void compile() throws GradingException {
         try {
+            gradingContext.observer().update("Verifying code...");
+
             StudentCodeReader reader = StudentCodeReader.from(gradingContext);
             for(StudentCodeInteractor verifier : currentVerifiers) verifier.interact(gradingContext, reader);
             for(StudentCodeInteractor modifier : currentModifiers) modifier.interact(gradingContext, reader);
+
             packageRepo();
         } catch (IOException e) {
             throw new GradingException("Failed to read project contents", e);
@@ -37,7 +40,7 @@ public class CompileHelper {
      * Packages the student repo into a jar
      */
     private void packageRepo() throws GradingException {
-        gradingContext.observer().update("Packaging repo...");
+        gradingContext.observer().update("Compiling code...");
 
         ProcessBuilder processBuilder = new ProcessBuilder();
         processBuilder.directory(gradingContext.stageRepo());
@@ -50,8 +53,6 @@ public class CompileHelper {
         } catch (ProcessUtils.ProcessException ex) {
             throw new GradingException("Failed to package repo", ex);
         }
-
-        gradingContext.observer().update("Successfully packaged repo");
     }
 
     /**
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -59,10 +59,10 @@ public class CompileHelper {
         try {
             ProcessUtils.ProcessOutput output = ProcessUtils.runProcess(processBuilder, 90000); //90 seconds
             if (output.statusCode() != 0) {
-                throw new GradingException("Failed to package repo: ", getMavenError(output.stdOut()));
+                throw new GradingException("Failed to compile", getMavenError(output.stdOut()));
             }
         } catch (ProcessUtils.ProcessException ex) {
-            throw new GradingException("Failed to package repo", ex);
+            throw new GradingException("Failed to compile: %s".formatted(ex.getMessage()), ex);
         }
     }
 
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -10,15 +10,19 @@ import java.util.List;
 public class CompileHelper {
     private final GradingContext gradingContext;
 
-    public CompileHelper(GradingContext gradingContext) {this.gradingContext = gradingContext;}
+    public CompileHelper(GradingContext gradingContext) {
+        this.gradingContext = gradingContext;
+    }
 
-    private final Collection<StudentCodeModifier> currentModifiers =
-            List.of(new ProjectStructureVerifier(), new PomModifier(), new PassoffJarModifier(), new TestFactoryModifier());
+    private final Collection<StudentCodeInteractor> currentVerifiers =
+            List.of(new ProjectStructureVerifier());
+
+    private final Collection<StudentCodeInteractor> currentModifiers =
+            List.of(new PomModifier(), new PassoffJarModifier(), new TestFactoryModifier());
 
     public void compile() throws GradingException {
-        for(StudentCodeModifier modifier : currentModifiers) {
-            modifier.modifyCode(gradingContext);
-        }
+        for(StudentCodeInteractor verifier : currentVerifiers) verifier.interact(gradingContext);
+        for(StudentCodeInteractor modifier : currentModifiers) modifier.interact(gradingContext);
         packageRepo();
     }
 
--- a/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
@@ -53,7 +53,7 @@ public class ModuleIndependenceVerifier implements StudentCodeVerifier {
             for (int i = 0; i < contents.size(); i++) {
                 String line = contents.get(i).trim();
                 if (line.isEmpty()) continue;
-                if (line.matches("^(?!import|package|//).*\\b(class|record|enum|@?interface)\\b")) break;
+                if (line.matches("^(/\\*\\*|@|(?!import|package|//).*\\b(class|record|enum|@?interface)\\b).*$")) break;
 
                 String packageImport = line.substring(line.indexOf(' ')).trim();
                 if (packageImport.startsWith("static")) {
--- a/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
@@ -56,7 +56,7 @@ public class ModuleIndependenceVerifier implements StudentCodeVerifier {
                 if (line.matches("^(?!import|package|//)(/\\*\\*|@|.*\\b(class|record|enum|@?interface)\\b).*$")) break;
 
                 String packageImport = getPackageImport(line);
-                if (packages.contains(packageImport)) {
+                if (packageImport != null && packages.contains(packageImport)) {
                     String warning = ("File %s imports from package %s (line %d), which exists in another module. " +
                             "The client and server modules should be independent")
                             .formatted(context.stageRepo().toPath().relativize(file.toPath()), packageImport, i + 1);
@@ -68,7 +68,7 @@ public class ModuleIndependenceVerifier implements StudentCodeVerifier {
 
     private static String getPackageImport(String line) {
         int firstSpace = line.indexOf(' ');
-        if(firstSpace == -1) return line;
+        if(firstSpace == -1) return null;
         String packageImport = line.substring(firstSpace).trim();
         int lastPeriod = packageImport.lastIndexOf('.');
         if(lastPeriod != -1) {
--- a/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
@@ -6,12 +6,10 @@ import edu.byu.cs.autograder.GradingException;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public class ModuleIndependenceVerifier implements StudentCodeInteractor {
     @Override
@@ -47,7 +45,7 @@ public class ModuleIndependenceVerifier implements StudentCodeInteractor {
         for(File file : files) {
             List<String> contents = reader.getFileContents(file);
             for (String line : contents) {
-                if (line.matches("public (?:(?:abstract)? class|interface|record) .* \\{")) break;
+                if (line.matches("(?:abstract )?class|interface|record")) break;
                 for (String classImport : classes) {
                     if (line.matches("import %s;".formatted(classImport))) {
                         String warning = "File %s imports %s. The client and server modules should be independent"
--- a/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
@@ -20,6 +20,11 @@ public class ModuleIndependenceVerifier implements StudentCodeVerifier {
             Set<File> clientFiles = reader.filesMatching(".*client/src/main/java/.*").collect(Collectors.toSet());
             Set<String> clientPackages = packageNames(clientFiles, reader);
 
+            Set<String> duplicatedPackages = new HashSet<>(serverPackages);
+            duplicatedPackages.removeIf(packageName -> !clientPackages.contains(packageName));
+            serverPackages.removeIf(duplicatedPackages::contains);
+            clientPackages.removeIf(duplicatedPackages::contains);
+
             checkImports(context, reader, serverFiles, clientPackages);
             checkImports(context, reader, clientFiles, serverPackages);
         } catch (IOException e) {
--- a/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
@@ -53,7 +53,7 @@ public class ModuleIndependenceVerifier implements StudentCodeVerifier {
             for (int i = 0; i < contents.size(); i++) {
                 String line = contents.get(i).trim();
                 if (line.isEmpty()) continue;
-                if (line.matches("^(/\\*\\*|@|(?!import|package|//).*\\b(class|record|enum|@?interface)\\b).*$")) break;
+                if (line.matches("^(?!import|package|//)(/\\*\\*|@|.*\\b(class|record|enum|@?interface)\\b).*$")) break;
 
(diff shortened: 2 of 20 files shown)


src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,59 @@
...
+            throw new GradingException("Unable to verify module independence", e);
+        }
+    }
+
+    private Set<String> qualifiedNames(Set<File> files, Path relativeTo) {
+        Set<String> classes = new HashSet<>();
+        for(File file : files) {
+            String fileName = relativeTo.relativize(file.toPath()).toString();
+            if(!fileName.contains(".java")) continue;
+            classes.add(fileName.substring(0, fileName.indexOf('.')).replace('/', '.'));
```
frozenfrank, line 38: It is really nice that Java requires this so we can read the class names out of the files!

What if a student has a `public` class nested inside another class?
1. If they used it properly within it's own module, would it be OK?
2. If they accessed it improperly from the wrong module, would this detect it?

If it passes (1) but fails (2), I think that's ok. I just can't fail (1).

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        return classes;
+    }
+
+    private void checkImports(GradingContext context, StudentCodeReader reader, Set<File> files, Set<String> classes)
+            throws IOException {
+        for(File file : files) {
+            List<String> contents = reader.getFileContents(file);
+            for (String line : contents) {
+                if (line.matches("(?:abstract )?class|interface|record")) break;
```
frozenfrank, line 48: What about `Enum` and `Annotation` type files?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,59 @@
...
+            if(!fileName.contains(".java")) continue;
+            classes.add(fileName.substring(0, fileName.indexOf('.')).replace('/', '.'));
+        }
+        return classes;
+    }
+
+    private void checkImports(GradingContext context, StudentCodeReader reader, Set<File> files, Set<String> classes)
+            throws IOException {
+        for(File file : files) {
+            List<String> contents = reader.getFileContents(file);
```
frozenfrank, line 46: What about files that are **not** `.java` files, like `.md`, `.sh`, `.txt`, `.gitignore` or otherwise? Will those kinds of files be excluded?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,59 @@
...
+            for (String line : contents) {
+                if (line.matches("(?:abstract )?class|interface|record")) break;
+                for (String classImport : classes) {
```
frozenfrank, line 49: Currently, you're iterating over all classes to see if any of them match the format. However, the format is fixed. It's always `import <CLASS_NAME>;` It would be a lot more effective to check for set inclusion by using a HashSet of all the known classes.

That makes this an N*M operation where N is the number of import lines and M is the number of classes. You could make it an N + M operation instead which would be basically processing each line & import once.

You could do either of the following to improve the efficiency:
1. Prepare a `HashSet` in advance with the line contents pre-appended, or
2. Extract the class name from the import statement and then compare it against the `HashSet`

src/main/java/edu/byu/cs/autograder/compile/ProjectStructureVerifier.java
```diff
@@ -4,17 +4,47 @@
...
+            default:
+                filePaths.add("shared/src/main/java");
+                filePaths.add("shared/src/test/java");
+                filePaths.add("server/src/main/java");
+                filePaths.add("client/src/main/java");
+        }
+
+        for(String filePath : filePaths) {
+            File file = new File(context.stageRepo(), filePath);
+            if(!file.exists() || !file.isDirectory()) {
```
frozenfrank, line 45: This looks good. Nice work!

src/main/java/edu/byu/cs/autograder/compile/StudentCodeReader.java
```diff
@@ -0,0 +1,50 @@
...
+        }
+        return files;
+    }
+
+    private StudentCodeReader(Set<File> files) {
+        this.files = files;
+    }
+
+    public List<String> getFileContents(File file) throws IOException {
+        if(!fileContents.containsKey(file)) {
```
frozenfrank, line 36: I like this caching of the file contents!

After reading in the entire repo, this would be holding a lot of data in memory. After the checks have stopped, is there a way for us to manually free the resources so that they aren't preserved through the whole grading process? I assume they're freed up at the end of the student's grading run, but this seems like something we would want to free up earlier if it's no longer needed.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
frozenfrank: Do you even need the non-capturing, optional match for `(?:abstract )?` clause?

frozenfrank (commented):
These changes look good, and I'm impressed with how quickly you solved this problem!

I have left a few comments in the PR about some optimizations, but otherwise, it's looking good to me.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: 1. Yes. This is only detecting what they do wrong, not what they do correctly.
2. I suppose not. For that matter, I wouldn't detect static imports either. Do you think that's a big deal?

src/main/java/edu/byu/cs/autograder/compile/StudentCodeReader.java
19mdavenport: Pretty sure the reader goes out of scope when the verification, modification, and compiling is over.

19mdavenport: But good point, It would probably be better to make it go out of scope even before we start compiling

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: Good points. Firstly, this would catch static imports. Do you think that's a big deal?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: Ok, We should be detecting static imports now. 1. is still always yes. 2 I'm unsure of atm

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: How about this: checking until we see anything that doesn't start with "package" or "import"?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: How does that look? I think I found a way to do both

src/main/java/edu/byu/cs/autograder/compile/StudentCodeReader.java
19mdavenport: Now it goes out of scope after validation, before we even start modifying anything

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: Looks like java doesn't allow you to import nested classes without importing the base containing class. Further more, only the nested class import is necessary. So the answers to both questions are yes

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
frozenfrank: It makes a ton of sense to have this optimization in place. Defining it is tricky though.

## Existing Approach

Here are several considerations for the existing approach:
* Sometimes there are blank lines between packages and imports, so we need to not let those trigger any early exits.
* There could exist comments in between imports (maybe some imports are commented out instead of removed, or maybe the student added comments for some reason) so we would need to skip those as well.
* I'm not sure if files are allowed to execute static blocks of code, and then import additional statements.
* It feels like a very safe assumption that all import statements precede the first class/record/enum/annotation. Java enforces that only one public object can exist in each file; additionally, it would just be really weird style to author an item, and then import additional items within the same file.

Therefore, it will be safe to keep the `line.matches() ... break;` statement with some modifications to the regex:
* Other items may _partially contain_ these words (i.e. "import package.MyClass;" or "import school.classroom;")

### Working Regex
The following regex will solve address the needs discussed so far. View it and several test cases on [regex101.com](https://regex101.com/r/ZWGoTq/1):
```regex
"^(?!import|\s*//).*\b(class|record|enum|annotation|interface)\b"
```

## Alternative Approach
If desired, another approach could be taken that always scans every line of the file. If the line does not start with "import ", then it is skipped, otherwise, the package name is extracted and compared against the set. This would be simpler, and assumption free. It's possible that the performance difference would never be noticed.

I think the existing approach is sufficient, and the assumptions made are good enough. However, another idea is presented for comparison's sake.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
frozenfrank: Great! Nice work testing that and determine that it always works!

src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
```diff
@@ -15,26 +15,37 @@ public CompileHelper(GradingContext gradingContext) {
...
-    private final Collection<StudentCodeInteractor> currentVerifiers =
+    private final Collection<StudentCodeVerifier> currentVerifiers =
             List.of(new ProjectStructureVerifier(), new ModuleIndependenceVerifier());
 
-    private final Collection<StudentCodeInteractor> currentModifiers =
+    private final Collection<StudentCodeModifier> currentModifiers =
             List.of(new PomModifier(), new PassoffJarModifier(), new TestFactoryModifier());
 
     public void compile() throws GradingException {
+        verify();
```
frozenfrank, line 25: Nice work splitting these separate responsibilities into separate functions. I assume that, besides the improvement in overall readability, this also helps the `StudentCodeReader` go out of scope properly?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: Blank lines are already preskipped. I was originally worried about having weird imports, but the I like the negative lookahead

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -5,53 +5,66 @@
...
 
-    private Set<String> qualifiedNames(Set<File> files, Path relativeTo) {
-        Set<String> classes = new HashSet<>();
+    private Set<String> packageNames(Set<File> files, StudentCodeReader reader) throws IOException {
+        Set<String> packages = new HashSet<>();
         for(File file : files) {
-            String fileName = relativeTo.relativize(file.toPath()).toString();
-            if(!fileName.contains(".java")) continue;
-            classes.add(fileName.substring(0, fileName.indexOf('.')).replace('/', '.'));
+            if(!file.getName().contains(".java")) continue;
```
frozenfrank, line 33: Did you mean to use `.endsWith(".java");` instead of `.contains()`? What examples of files would you want to include that contain ".java" in the middle, but DO NOT end with ".java"?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -5,53 +5,66 @@
...
+            List<String> contents = reader.getFileContents(file);
+            if(contents.isEmpty()) continue;
+            String firstLine = contents.getFirst();
+            if(firstLine.startsWith("package")) {
+                packages.add(firstLine.substring(8, firstLine.indexOf(';')));
```
frozenfrank, line 38: Is it possible for a student to create a **client package** named "MyPackage" and separately create a **server package** also named "MyPackage"? Could they do that with classes as well?

One approach would be to qualify our warnings by first detecting that it exists in the other module, and then excluding it if it is defined in the current module.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -5,53 +5,66 @@
...
-                for (String classImport : classes) {
-                    if (line.matches("import %s;".formatted(classImport))) {
-                        String warning = "File %s imports %s. The client and server modules should be independent"
-                                .formatted(context.stageRepo().toPath().relativize(file.toPath()), classImport);
-                        context.observer().notifyWarning(warning);
-                    }
+            for (int i = 0; i < contents.size(); i++) {
+                String line = contents.get(i).trim();
+                if (line.isEmpty()) continue;
+                if (!(line.startsWith("package") || line.startsWith("import"))) {
```
frozenfrank, line 51: What do you think about using [this regex](https://regex101.com/r/ZWGoTq/4) to determine when we should `break`?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -5,53 +5,66 @@
...
+                    break;
+                }
+
+                String packageImport = line.substring(line.indexOf(' ')).trim();
+                if (packageImport.startsWith("static")) {
+                    packageImport = packageImport.substring(packageImport.indexOf(' '), packageImport.lastIndexOf('.')).trim();
```
frozenfrank, line 57: I appreciate this intentional string cleaning before reporting it to the user :)

frozenfrank (commented):
The is looking good, and several of my original concerns were addressed.

I do have a few new questions, and I put together [a cool regex with test suites](https://regex101.com/r/ZWGoTq/4) to verify that it works in the cases we care about. You should be able to "update" the regex with new test cases and improved detection within Regex101.com if you find additional ways to improve it.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,72 @@
...
+                }
+                if (packageImport.contains(".")) {
+                    packageImport = packageImport.substring(0, packageImport.lastIndexOf('.'));
+                }
+
+                if (packages.contains(packageImport)) {
```
frozenfrank, line 63: I actually missed this the first time through because it was small. This looks great!

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
frozenfrank: > `if (packages.contains(packageImport))`

I think it's perfect 👌

frozenfrank:
@19mdavenport Can I help carry this PR across the finish line now that we've made it this far? At this point, I feel like the main points have been discussed— or at least raised.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: Good point. How does it look now?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
19mdavenport: Yes. Thank you

Fiwafoofa:
When I ran this against my code, I get a index error within my SqlDAO.java file [here](https://github.com/Fiwafoofa/chessTA/blob/removed-modified-test-files/server/src/main/java/dataaccess/mysql/SqlDAO.java) on (I think) line 18 with the curly brace when executing the `ModuleIndependenceVerifier` with a phase 3 submission. Line 58 `String packageImport = line.substring(line.indexOf(' ')).trim();` caused the error with `line` being equal to `"}"`.  My guess is something was off with the Regex. I have to go to class, but when I get back I'll do more debugging.

Screenshot 2024-05-17 095854

19mdavenport:
@Fiwafoofa thank you for looking into that. I think it might just be the beginning string marker (^) at the beginning. I'll try removing that and seeing if that helps

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -69,4 +65,19 @@ private void checkImports(GradingContext context, StudentCodeReader reader, Set<
             }
         }
     }
+
+    private static String getPackageImport(String line) {
+        int firstSpace = line.indexOf(' ');
+        if(firstSpace == -1) return line;
```
frozenfrank, line 71: It's interesting that this line still returns the line for comparison against the set of packages. I suppose it will never succeed, and the comparison isn't horribly expensive because of the use of `HashSet`'s, but I would expect that if we weren't able to detect any packages in this line, that the line would be skipped and no further processing would occur.

Additionally, for those other developers who aren't intimately familiar with the problem going on here, could you provide a JavaDoc comment or even any comment that provides an overview of the kinds of patterns this is intended to match?

Otherwise, I love how carefully you've processed all this information to parse out the data we're interested in!

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
frozenfrank: Nice job handling those cases!

frozenfrank (approved):
This resolves all of my major concerns. There are still other adjustments I would make to improve the readability and maintainability of the code going forward, but otherwise, you've designed it well.

I do worry that we're trying too hard to optimize the evaluation of the packages. If the regex becomes much more complex than it already is, it may be better to just read all the lines, but only try to compare those that start with "import". Very quickly, we're seeing that a lot of corner cases exist that will cause the code to behave it ways we didn't intend. Even if they're not mainstream, someone is bound to trigger them. At the present, I think we're OK, but I don't want to go much farther with our assumptions and our specific exclusions than we already have.

I would want @Fiwafoofa to test it again and see how the adjustments work on his repo.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -20,6 +20,11 @@ public void verify(GradingContext context, StudentCodeReader reader) throws Grad
             Set<File> clientFiles = reader.filesMatching(".*client/src/main/java/.*").collect(Collectors.toSet());
             Set<String> clientPackages = packageNames(clientFiles, reader);
 
+            Set<String> duplicatedPackages = new HashSet<>(serverPackages);
+            duplicatedPackages.removeIf(packageName -> !clientPackages.contains(packageName));
+            serverPackages.removeIf(duplicatedPackages::contains);
```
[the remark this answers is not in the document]
19mdavenport: I still don't think you do understand what they're doing. Look at line 24.

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -48,18 +53,9 @@ private void checkImports(GradingContext context, StudentCodeReader reader, Set<
...
-                }
-
-                String packageImport = line.substring(line.indexOf(' ')).trim();
-                if (packageImport.startsWith("static")) {
-                    packageImport = packageImport.substring(packageImport.indexOf(' '), packageImport.lastIndexOf('.')).trim();
-                }
-                if (packageImport.contains(".")) {
-                    packageImport = packageImport.substring(0, packageImport.lastIndexOf('.'));
-                }
+                if (line.matches("^(/\\*\\*|@|(?!import|package|//).*\\b(class|record|enum|@?interface)\\b).*$")) break;
```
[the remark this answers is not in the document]
19mdavenport: 1. These seem equally clear to me, but if putting it the negative lookahead first would be clearer to you I'm amenable to that.
2. You would think that, but empirically it does actually matter. Either 1-Regex101 is incorrect, 2-Something changed in java between Java 8 (Regex101) and Java 21 (autograder) that matters here, or 3-something is different about my implementation and the test cases on Regex101

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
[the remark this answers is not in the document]
19mdavenport: Does https://github.com/softwareconstruction240/autograder/pull/360/commits/d9d827da4e3fa73829a344e5c3fa78a35142d029 make it clearer?

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Gets just the package being imported from a java import statement
+     * <p>For example, "import java.util.Set;" would return "java.util"
+     * <p>For example, "import static java.sql.Types.NULL;" would return "java.sql"
+     *
+     * @param line java import statement
+     * @return package name
+     */
+    private static String getPackageImport(String line) {
+        int firstSpace = line.indexOf(' ');
```
frozenfrank, line 76: Per a discussion with @19mdavenport, let's use [this REGEX](https://regex101.com/r/rSSd12) to verify that the line is a valid import line. If it's not, we'll `return null`, signifying that this shouldn't be searched.

This additional check is intended to be more precise than the basic but fast check for `line.startsWith("import")` that is called before evaluating this method. Honestly, it probably makes sense to put those two statements next to each other for clarity, but that decision is left up to the implementer :)

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        return packageImport;
+    }
+
+    /**
+     * Removes items common in all provided collections
+     *
+     * @param collections variable number of collections
+     */
+    private void removeCommonItems(Collection<?>... collections) {
```
frozenfrank, line 95: Nice work with the abstract collection of collections!

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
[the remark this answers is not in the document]
frozenfrank: Yes, thank you!

src/main/java/edu/byu/cs/autograder/compile/ModuleIndependenceVerifier.java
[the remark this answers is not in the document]
frozenfrank: I now understand that in Java, the regex matcher is looking to match the entire line. Since that's the case, it is necessary to include the `.*$` which also clarifies the role of the entire regex.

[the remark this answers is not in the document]
frozenfrank: In a discussion with @19mdavenport, we decided to migrate away from this "shortcut" approach because it has the possibility of many edge cases that could misbehave. Instead, we'll scan the whole file, but only pay attention to lines that, when trimmed, begin with "import" and then attempt to extract the import statement.

PR #369, opened 2024-05-11
Disable canvas and compilation

frozenfrank:
This PR continues efforts begun in #328 by adding guards to newly added method calls. It also leverages the newly added `ApplicationProperties` system to read in the values from a properties file.

### DIFF
--- a/src/main/java/edu/byu/cs/server/Server.java
+++ b/src/main/java/edu/byu/cs/server/Server.java
@@ -151,6 +151,9 @@ public class Server {
             if (cmd.hasOption("use-canvas")) {
                 properties.setProperty("use-canvas", cmd.getOptionValue("use-canvas"));
             }
+            if (cmd.hasOption("disable-compilation")) {
+                properties.setProperty("run-compilation", "false");
+            }
         } catch (ParseException e) {
             throw new RuntimeException("Error parsing command line arguments", e);
         }
@@ -169,6 +172,7 @@ public class Server {
         options.addOption(null, "cas-callback-url", true, "CAS Callback URL");
         options.addOption(null, "canvas-token", true, "Canvas Token");
         options.addOption(null, "use-canvas", true, "Using Canvas");
+        options.addOption(null, "disable-compilation", false, "Turn off student code compilation");
         return options;
     }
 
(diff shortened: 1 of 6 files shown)


frozenfrank:
Note that the commit message "RUN_COMPILATION can be controlled via ApplicationProperties" in commit 50cb2c6 is mostly a theoretical question. Nobody has shown me nor demonstrated where the `.properties` file exists so I'm not sure how to actually use it.

19mdavenport:
There is no .properties file associated with ApplicationProperties. It's tied to command line arguments

frozenfrank:
> There is no .properties file associated with ApplicationProperties. It's tied to command line arguments

That explains it. I guess I just need some training on how to use the commandline arguments to adjust the values.

Is there something more you're looking for from this PR, or how can I help this one become approved as well?

19mdavenport (changes requested):
Could you add support for your new property to where it exists in main Server class, as well as providing it as an option in the docker compose file?

frozenfrank:
> Could you add support for your new property to where it exists in main Server class, as well as providing it as an option in the docker compose file?

@19mdavenport I can look into the main `Server` class, but I have now idea how Docker works or while file I would be manipulating. I'll do my best.

19mdavenport:
https://github.com/softwareconstruction240/autograder/blob/main/compose.yml you'd just need to add it with the rest of the command line arguments

frozenfrank (commented):
@19mdavenport I've added the support you asked for. 

(I still don't understand everything in that file, but I was able to follow the pattern to add support for the property. Thanks for the tip!)

src/main/java/edu/byu/cs/server/Server.java
```diff
@@ -151,6 +151,9 @@ private static void setupProperties(String[] args) {
             if (cmd.hasOption("use-canvas")) {
                 properties.setProperty("use-canvas", cmd.getOptionValue("use-canvas"));
             }
+            if (cmd.hasOption("disable-compilation")) {
+                properties.setProperty("run-compilation", "false");
+            }
```
19mdavenport, line 156: Nice with the optional parameter

19mdavenport (approved):

PR #371, opened 2024-05-14
Only count completed extra credit in notes

19mdavenport:
This fixes a calculation where students are receiving notes like "Extra credit tests: +5.6%" (had done 1/2 extra credit) and "Extra credit tests: +2.4%" (had done no extra credit). This will be replaced with the correct number (no extra credit display/4%/8%)

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/test/PassoffTestGrader.java
+++ b/src/main/java/edu/byu/cs/autograder/test/PassoffTestGrader.java
@@ -75,7 +75,7 @@ public class PassoffTestGrader extends TestGrader {
 
         Map<String, Float> ecScores = getECScores(testResults);
         float extraCreditValue = PhaseUtils.extraCreditValue(gradingContext.phase());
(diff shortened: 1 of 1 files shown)


frozenfrank (approved):
This adjustment seems reasonable to me.

I'm having hard time figuring our what was actually wrong with the previous version. The lack of documentation on surrounding methods makes it quite difficult to figure out what is supposed to happen at each stage of the process. However, I believe that this works.

Was the previous code generating floating point rounding errors, which lead to the strange values? What is the extra portion in the `ecScores` that we need to `floor` off before adding it?

19mdavenport:
The value was the percentage of passed extra credit tests. Some extra credit tests are passed by all students who don't attempt the extra credit as they check that valid moves don't contain what would be valid extra credit moves under different circumstances. It was displaying to students that they had gotten +2.4% or +5.6% extra credit, when the correct values are discreet: 0%, 4%, 8%

PR #374, opened 2024-05-18
Suppress expected warning output in tests

frozenfrank:
Previously, the `LateDayCalculatorTest`'s would generate error output as they intentionally fed poorly formatted data to the functions and tested the results. While we do intend to warn developers as they work on these kinds of adjustments, it is not good practice to have error output hitting the console when the tests run.

The PR introduces a new quiet mode that addresses this problem. The tests use overloads of the methods that enable them to disable the warning output, but the production code still runs with the output enabled.

### DIFF
--- a/src/test/java/edu/byu/cs/autograder/score/LateDayCalculatorTest.java
+++ b/src/test/java/edu/byu/cs/autograder/score/LateDayCalculatorTest.java
@@ -10,6 +10,14 @@ import java.time.format.DateTimeFormatter;
 
 class LateDayCalculatorTest {
 
+    /**
+     * To be enabled in production.
+     * Removes the warning output since the tests are expected to pass in faulty input values.
+     * <br>
+     * Disable during debugging.
+     */
+    private static final boolean QUIET_HOLIDAY_INIT_WARNINGS = true;
+
     @Test
     void getNumDaysLateWithoutHolidays() {
         // Initialize without holidays
@@ -70,7 +78,7 @@ class LateDayCalculatorTest {
     void getNumDaysLateWithHolidays() {
         // Initialize with holidays
         LateDayCalculator standardlateDayCalculator = new LateDayCalculator();
-        standardlateDayCalculator.initializePublicHolidays(getMultilinePublicHolidaysConfiguration());
+        standardlateDayCalculator.initializePublicHolidays(getMultilinePublicHolidaysConfiguration(), QUIET_HOLIDAY_INIT_WARNINGS);
 
         // See image: days-late-with-holidays-common
         String commonDueDate = "2024-03-07 11:59:00 PM -07:00";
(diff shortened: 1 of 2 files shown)


src/test/java/edu/byu/cs/autograder/score/LateDayCalculatorTest.java
```diff
@@ -10,6 +10,14 @@
 
 class LateDayCalculatorTest {
 
+    /**
+     * To be enabled in production.
+     * Removes the warning output since the tests are expected to pass in faulty input values.
+     * <br>
+     * Disable during debugging.
+     */
+    private static final boolean QUIET_HOLIDAY_INIT_WARNINGS = true;
```
19mdavenport, line 19: Would we want the default value in the repo to be false?

frozenfrank: Well, it depends on how you define "default value in the repo."

## What Should Happen Generally?
It is factual that we do **not** want to suppress the warnings by default most of the time. Most of the time, we want `quietWarnings` set to false inside `LateDayCalculator.java`. This is accomplished by providing simple overloads of all the relevant methods which do not accept any boolean value, and then pass through the value `false` to the implementing method of `LateDayCalculator#initializePublicHolidays()`. This essentially makes the argument optional, since the overload without it is available for all different call signatures, and makes enables the warnings most of the time.

This is a snippet of the code from this PR that accomplishes this effect:
```java
// src/main/java/edu/byu/cs/autograder/score/LateDayCalculator.java

public Set<LocalDate> initializePublicHolidays(@NonNull String encodedPublicHolidays) {
    return initializePublicHolidays(encodedPublicHolidays, false);
}
...
public Set<LocalDate> initializePublicHolidays(@NonNull String encodedPublicHolidays, @NonNull String dateEncodingFormat) {
    return initializePublicHolidays(encodedPublicHolidays, dateEncodingFormat, false);
}
```

## What Should Happen Here?
It sounds like you're suggesting that this line is the "default value for the repo," and that this line should have the value `false`.

Specifically in this LateDayCalculator**Test**.java file, we do want to suppress the warnings. That's the whole point of this PR. This `private static ...` variable is only used within the test file to turn on or off the suppression for the test cases collectively. If a developer needs to debug any of the test cases, they would turn this value to `false` temporarily, but we want the suppression enabled for the tests most of the time. If a developer wants to enable debugging for any test case specifically, the developer would replace the variable reference with a boolean literal in the appropriate place.

## Therefore, what?
This PR is designed and built the way it was intended. It's already passed the tests. I recommend:
1. Using the PR button to rebase it onto the latest
2. Allowing the tests to pass again
3. Merge into `main`

19mdavenport: Sorry. It looked like that variable was doing the opposite of what it is. I thought that true meant that the warnings would display, not that they would be prevented from displaying. This looks good.

19mdavenport (approved):

PR #380, opened 2024-05-29
378 get all submission endpoint giving 500

19mdavenport:
I fixed this by creating a central Serializer class to run all json serialization through


frozenfrank (commented):
Yup, looks good!

PR #381, opened 2024-05-31
Modified test files ignore phase quality

Fiwafoofa:
Quick bug fix for not running the modified test files verifier on code quality and making a helper method throw a runtime exception instead of returning null.

### DIFF
--- a/src/main/java/edu/byu/cs/util/FileUtils.java
+++ b/src/main/java/edu/byu/cs/util/FileUtils.java
@@ -200,7 +200,9 @@ public class FileUtils {
                 }
             }
         } catch (IOException e) {
-            return null;
+            throw new RuntimeException(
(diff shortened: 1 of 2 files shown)


src/main/java/edu/byu/cs/util/FileUtils.java
```diff
@@ -200,7 +200,9 @@ public static Map<String, String> getFileNamesToAbsolutePaths(Path filePath) {
                 }
             }
         } catch (IOException e) {
-            return null;
+            throw new RuntimeException(
+                    String.format("Could not find file names given %s: %s", filePath, e.getMessage())
+            );
```
19mdavenport, line 205: Would you be willing to change this to be a GradingException?

19mdavenport (changes requested):

19mdavenport (approved):

PR #239, opened 2024-03-14
228 Approve blocked submissions

DallinFromEarth:
This PR will build off of the work that is being done in #233

Once the code written in #233 has marked the new `approved` variable as `false` on a submission, this PR will do the following:
- Mark the submission in the admin view as needing approval (both in the submissions tab, and on the student's info box)
- Provide an interface for any admin to approve the submission and send the grade to canvas
	- They will have the option to approve with penalty (currently set to 10%)
	- Or to approve without penalty (meaning they get full points)
- Notify the autograder server of the admin's decision
- Update the autograder database of the approved status and score
- Send the score to canvas with a note

## Database Modifications
Run the following SQL commands to configure the database with the needed values:
```sql
INSERT INTO autograder.rubric_config (phase, type, category, criteria, points) VALUES ('Commits', 'GIT_COMMITS', 'Git Commits', 'At least 10 GitHub commits evenly spread over the assignment period that demonstrate proof of work.', 0);
```

PR #383, opened 2024-06-03
300 fullstack refactor rubicitems to be dynamic

19mdavenport:
Should be fairly self explanatory. Converted the three RubricItem's to an EnumMap, so we can add things to the enum and it becomes a new item


19mdavenport:
@frozenfrank These changes may be helpful with #239

src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -202,9 +201,11 @@ private boolean passed(Rubric rubric) {
         boolean passed = true;
 
         boolean isPassoffRequired = PhaseUtils.isPassoffRequired(gradingContext.phase());
-        if (isPassoffRequired && rubric.passoffTests() != null && rubric.passoffTests().results() != null)
-            if (rubric.passoffTests().results().score() < rubric.passoffTests().results().possiblePoints())
-                passed = false;
+        var passoffTestItem = rubric.items().get(Rubric.RubricType.PASSOFF_TESTS);
+        if (isPassoffRequired && passoffTestItem != null && passoffTestItem.results() != null &&
```
frozenfrank, line 205: I know that the single-exit point execution is sometimes desirable, but I think this code would be a lot more readable with guarded exits.

I.E.
```java
if (!isPassoffRequired) return true;
...
if (passoffTestItem == null || passoffTestItem.results() == null) return true;
...
return passoffTestItem.results().score() >= passoffTestItem.results().possiblePoints();
```

src/main/java/edu/byu/cs/autograder/test/PreviousPhasePassoffTestGrader.java
```diff
@@ -47,7 +48,7 @@ private <T> Set<T> allPreviousPhases(PhaseFunction<T> func) throws GradingExcept
             Phase previous = gradingContext.phase();
             while ((previous = PhaseUtils.getPreviousPhase(previous)) != null) {
                 RubricConfig rubricConfig = DaoService.getRubricConfigDao().getRubricConfig(previous);
-                if(rubricConfig.passoffTests() != null) {
+                if(rubricConfig.items().get(Rubric.RubricType.PASSOFF_TESTS) != null) {
```
frozenfrank, line 51: Note: `RubricConfig` is still a class. We could provide helpful methods to preserve the existing method calls, and avoid all these code updates.

src/main/java/edu/byu/cs/canvas/CanvasUtils.java
```diff
@@ -26,47 +27,24 @@ public class CanvasUtils {
      */
     public static Rubric decimalScoreToPoints(Phase phase, Rubric rubric) throws GradingException, DataAccessException {
         RubricConfig rubricConfig = DaoService.getRubricConfigDao().getRubricConfig(phase);
+        EnumMap<Rubric.RubricType, Rubric.RubricItem> rubricItems = new EnumMap<>(Rubric.RubricType.class);
 
-        Rubric.RubricItem convertedPassoffTests = null;
-        Rubric.RubricItem convertedUnitTests = null;
-        Rubric.RubricItem convertedQuality = null;
+        for(Rubric.RubricType type : Rubric.RubricType.values()) {
```
frozenfrank, line 32: It would be helpful to provide an iterating method in `RubricType` that calls a lambda on each of the values with the  `type` and the `rubricItem`. It would simplify this common iterating pattern used in multiple places...

src/main/resources/frontend/src/types/types.ts
```diff
@@ -40,7 +40,17 @@ export type RubricItem = {
     results: RubricItemResults,
 }
 
+//unfortunately with the way gson serializes maps and the way javascript parses that json each needs to be its own item
+export type RubricItems = {
```
frozenfrank, line 44: Consider using the builtin Typescript `Record` type to represent this more cleanly. If we already had a `type` that represents the different kinds of Phases, using that would be ideal, but I'm not sure if/where it exists.

```ts
export type RubricItems = Record<"PASSOFF_TESTS"|"UNIT_TESTS"|"QUALITY"|"GIT_COMMITS", RubricItem>;
export type RubricItems = Record<PhaseNames, RubricItem>; // Preferred, if available
```

src/main/resources/frontend/src/views/StudentView/ResultsPreview.vue
```diff
@@ -34,9 +35,9 @@ const openDetails = ref<boolean>(false);
     </div>
     <p class="submission-notes" v-html="submission.notes.replace('\n', '<br />')"></p>
     <div class="rubric-item-summaries">
-      <p v-if="submission.rubric.passoffTests">Functionality: {{ roundTwoDecimals(submission.rubric.passoffTests.results.score) }} / {{ submission.rubric.passoffTests.results.possiblePoints }}</p>
-      <p v-if="submission.rubric.quality">Code Quality: {{ roundTwoDecimals(submission.rubric.quality.results.score) }} / {{ submission.rubric.quality.results.possiblePoints }}</p>
-      <p v-if="submission.rubric.unitTests">Unit Tests: {{ roundTwoDecimals(submission.rubric.unitTests.results.score) }} / {{ submission.rubric.unitTests.results.possiblePoints }}</p>
+      <p v-if="getPassoffTests(submission.rubric)">Functionality: {{ roundTwoDecimals(getPassoffTests(submission.rubric).results.score) }} / {{ getPassoffTests(submission.rubric).results.possiblePoints }}</p>
```
frozenfrank, line 38: Can we replace this with a dynamic data structure as well? It seems the only additional piece we'd need to store would be the human readable label ("Functionality", "Code Quality", "Unit Tests", etc...), and then we could streamline these declarations with a kind of iterator instead!

src/main/resources/frontend/src/views/StudentView/SubmissionInfo.vue
```diff
@@ -47,16 +48,16 @@ const { submission } = defineProps<{
   </div>
   <div class="container">
     <RubricItemView
-      v-if="submission.rubric.passoffTests"
```
frozenfrank, line 50: ^Ditto

Vue's version of:
- For each rubric item (returned in our preferred order, ideally)
    - If the item value is non-null,
    - Render a `RubricItemView` with the item passed in properly

src/test/java/edu/byu/cs/autograder/score/ScorerTest.java
```diff
@@ -201,7 +199,9 @@ void score__phaseNotGradeable() {
                 mockObserver, false);
         addQueueItem(new QueueItem("testNetId", Phase.Phase0, Instant.now(), true));
 
-        Rubric emptyRubric = new Rubric(null, null, null, true, "testNotes");
+        Rubric emptyRubric = new Rubric(new EnumMap<>(Rubric.RubricType.class),
```
frozenfrank, line 202: I've seen this constructor called multiple times, and the TODO commend and empty parameters seem duplicated to me. How would you feel about introducing a specialized static constructor instead?

I.E.
```java
// rubric.java
  public static Rubric withRubricType(Class<Rubric.RubricType> rubricType, boolean passed, string notes) {
    return new Rubric(new EnumMap<>(rubricType),
      null, null, null, //TODO: Remove these between semesters
      passed, notes);
  }

// Other files, like this one
Rubric emptyRubric = Rubric.withRubricType(Rubric.RubricType.class, true, "testNotes");
```

frozenfrank (commented):
This looks like an overall major improvement to the autograder. I've left comments about potential improvements, but none of them are showstoppers. Thanks for making these helpful adjustments!

src/main/resources/frontend/src/types/types.ts
19mdavenport: Thank you. That provides the exact same functionality I attempted and failed to reproduce with a map

src/main/java/edu/byu/cs/autograder/test/PreviousPhasePassoffTestGrader.java
19mdavenport: There are only two instances where we get this with a specific type in mind rather than a dynamic variable type. I would think that adding extra methods would add extra clutter for not that much added benefit. Do you disagree?

src/main/java/edu/byu/cs/canvas/CanvasUtils.java
19mdavenport: That sounds like it would add a bunch of unnecessary code clutter without a ton of added benefit. We would have to add a new functional interface to make that lambda, unless you know of one that already exists and can throw custom exceptions.

src/test/java/edu/byu/cs/autograder/score/ScorerTest.java
19mdavenport: I don't think it even needs to be a static factory method. Can just be a constructor matching what the normal constructor will look like after spring. Good idea, thank you.

src/main/java/edu/byu/cs/autograder/score/Scorer.java
19mdavenport: Seems reasonable. How does that look?

Issue #376, opened 2024-05-21
Vary Git Commit Requirements by Phase


frozenfrank:
## Overview
The code required for each phase really is quite different, and it would make a lot sense to vary the requirements to match this. Specifically, the earliest phases do not really require 10 commits as they can be finished with 8 reasonably sized commits. TA's in the 2024 spring semester estimate that about 10% of students were flagged for having 8-9 commits even though the student's had a done a thorough job on the assignment. Those students received full credit.

Note that students receive essentially a free commit when they originally create their GitHub repo which create a commit. Usually, this occurs much earlier in the semester.

## Current Schedule
Currently require **10 commits** spaced over **3 days**, each with at least **5 line changes** for **all phases**.

## Proposed Schedule
The following table proposes a requirement schedule for all of the phases. While we could vary the number for each phase, I think it makes sense to break it into an intro period with lower requirements, a phase without any code, and then the implementation period with higher requirements. These requirements aren't designed to actually be difficult and all students should be easily meeting these requirements just by committing at reasonable checkpoints.

| Phase | Commits | Days | Line Changes |
|--------|--------|--------|--------|
| 0 | 8 | 2 | 5 |
| 1 | 8 | 2 | 5 |
| 3 | 12 | 3 | 5 |
| 4 | 12 | 3 | 5 |
| 5 | 12 | 3 | 5 | 
| 6 | 12 | 3 | 5 | 

Notice the increase in the line-change requirements. 20 line changes can be viewed as 10 lines modified (insertions + deletions), and I think it's OK to encourage students to write substantial commits rather than committing every single time they change any line.

This proposal is not final, and should be edited to reflect any agreed upon requirements.

## Todo
- [ ] Agree on a requirement schedule
- [ ] Add support for configuring these values based on the phase
- [ ] Update the phase spec grading tables to show the requirements
    - [ ] [Phase 0](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/0-chess-moves/chess-moves.md#grading-rubric)
    - [ ] [Phase 1](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/1-chess-game/chess-game.md#grading-rubric)
    - [ ] [Phase 3](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/3-web-api/web-api.md#grading-rubric)
[...]

PR #385, opened 2024-06-05
Extend the grading window by a configurable amount of minutes for grace

frozenfrank:
## Motivation
This implements the request made by @19mdavenport regarding a grace period:

> Currently the commit verification is finding the bounds on time between the last submission and when the current submission was made. Do you see any potential problems if we gave them an extra five minutes on either side to account for differences in computer times?
>
> A student got flagged for submitting 21 seconds earlier than the latest commit time

## Resolution
The fix is pretty straightforward: push forward the limit window by a small amount to allow for local/server clock inconsistencies. This introduces a configurable value to represent the exact amount, and then adjusts the limit. Thankfully, the change only needs to be made in a single place because the code is well organized.


19mdavenport (approved):

frozenfrank:
Thanks for reviewing it so quickly! I have another PR coming soon that addresses #376.

19mdavenport:
> Thanks for reviewing it so quickly!

Sure, honestly I didn't know how much you would be doing in the near future and thought I would have to make this change myself. This is very similar to what I was already thinking of doing.

PR #386, opened 2024-06-05
376 vary git commit requirements by phase

### LINKED ISSUE #376: Vary Git Commit Requirements by Phase

frozenfrank:
Resolves #376. Includes tests for #385.


src/main/java/edu/byu/cs/util/PhaseUtils.java
```diff
@@ -187,6 +188,19 @@ public static float extraCreditValue(Phase phase) {
...
+        var penaltyPct = 10;
+        var forgivenessMinutesHead = 3;
+        return new CommitVerificationConfig(10, 3, 5, penaltyPct, forgivenessMinutesHead);
+        // TODO: Enable this functionality to change the values in next semester.
+        // Don't forget to also update the rubrics in the phase specs and the grade-book.
+//        return switch (phase) {
+//            case Phase0, Phase1 -> new CommitVerificationConfig(8, 2, 5, penaltyPct, forgivenessMinutesHead);
+//            case Phase3, Phase4, Phase5, Phase6 -> new CommitVerificationConfig(12, 3, 5, penaltyPct, forgivenessMinutesHead);
+//            case Quality -> throw new GradingException("No commit verification for this phase");
+//        };
```
19mdavenport, line 201: Thank you for leaving this as this semester's requirements and a TODO comment with next semester's

src/main/java/edu/byu/cs/util/PhaseUtils.java
```diff
@@ -187,6 +188,19 @@ public static float extraCreditValue(Phase phase) {
...
 
+    public static CommitVerificationConfig verificationConfig(Phase phase) throws GradingException {
+        var penaltyPct = 10;
+        var forgivenessMinutesHead = 3;
+        return new CommitVerificationConfig(10, 3, 5, penaltyPct, forgivenessMinutesHead);
+        // TODO: Enable this functionality to change the values in next semester.
+        // Don't forget to also update the rubrics in the phase specs and the grade-book.
+//        return switch (phase) {
+//            case Phase0, Phase1 -> new CommitVerificationConfig(8, 2, 5, penaltyPct, forgivenessMinutesHead);
+//            case Phase3, Phase4, Phase5, Phase6 -> new CommitVerificationConfig(12, 3, 5, penaltyPct, forgivenessMinutesHead);
```
19mdavenport, line 199: Should the minimum number of lines changed be a variable as well? Also, would minimum lines, penalty percentage, and forgiveness minutes be better as class constants rather than local magic numbers?

src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -65,16 +66,11 @@ public Grader(String repoUrl, String netId, GradingObserver observer, Phase phas
         File stageRepo = new File(stagePath, "repo");
 
         // Init Grading Context
-        int requiredCommits = 10;
-        int requiredDaysWithCommits = 3;
-        int commitVerificationPenaltyPct = 10;
-        int minimumChangedLinesPerCommit = 5;
-
+        var cvConfig = PhaseUtils.verificationConfig(phase);
```
19mdavenport, line 69: Do you think this would be easier to read without using `var`?

src/main/java/edu/byu/cs/util/PhaseUtils.java
frozenfrank: I'll move it to a named variable as requested.

In truth, there are a wide range of different formats this code could take; it could quickly become a game of preferences to determine what should be named and what should be a number. Intelli-J inserts the names of the record fields inline when viewing it in an IDE, but this effect obviously won't be observed in the GitHub PR.

None of these are "magic numbers" because they aren't numbers in-lined into code expressions. These are named configuration variables that capable of being adjusted dynamically. The `CommitVerificationConfig` record clearly defines each one and provides information about what they represent. In the functioning code, these are referenced by name. This is the place where they are defined which is consistent with the definition of other fields and strings relating to Phases in this file.

However, this needn't be the place where they are defined. We could just as easily read these numbers out of a configurable table, and the system could handle it. In fact, these numbers have been extracted into this format specifically to simplify and enable the connection to a configurable table in the future. In this way the `CommitVerificationConfig` serves as an interface that must be implemented, but the data can be provided from any of many places.

This is the same reason they won't be replaced with class constants. Class constants cannot be configured remotely, but these values can be configured remotely.

src/main/java/edu/byu/cs/autograder/Grader.java
frozenfrank: Sure, that's fine.

I am used to relying on the inferred typing information more frequently, but it's OK to have it explicitly defined.

frozenfrank: After looking into it more, I see why I didn't include the type explicitly here. `PhaseUtils.verificationConfig()` returns a specific type, and then this type is directly passed through into `GradingContext`. If the type were to change, I wouldn't need to change this line at all, though I would change the place it's generated and the place it's used.

Since there's no explicit need to have the variable, typing it as `var` makes it more versatile to changes.

As a matter of face, I think I'll just in-line the reference and avoid making a variable in the first place.

frozenfrank: Except, that having the explicit type information emphasizes the ability of future developers to get that same type from some source other than PhaseUtils. It doesn't have to come from there, and they may combine config values from different places or different fields in order to construct the actual result.

I'll include the type information.

19mdavenport (approved):

PR #388, opened 2024-06-07
Add verification of running with port argument 0

19mdavenport:
This adds another verifier that checks for students hardcoding port 8080 and seeing if it can find a line of code matching `port = server.run(0)`. It then warns the student if it can find 8080 somewhere or can't find the target string

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/compile/verifers/ServerFacadeTestPortVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/verifers/ServerFacadeTestPortVerifier.java
@@ -1,0 +1,53 @@
+package edu.byu.cs.autograder.compile.verifers;
+
+import edu.byu.cs.autograder.GradingContext;
+import edu.byu.cs.autograder.GradingException;
+import edu.byu.cs.autograder.compile.StudentCodeReader;
+import edu.byu.cs.autograder.compile.StudentCodeVerifier;
+import edu.byu.cs.model.Phase;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ServerFacadeTestPortVerifier implements StudentCodeVerifier {
+    private static final String DYNAMIC_PORT_LINE = "port = server.run(0)";
+    private static final int CHECK_HARDCODED_PORT = 8080;
+    private static final String EXPLANATION = "A port argument of 0 lets Spark/Jetty choose a port. The run method " +
+            "then returns the port the server started on. Use this return value to set up your server facade " +
+            "(probably \"var port\").";
+
+    @Override
+    public void verify(GradingContext context, StudentCodeReader reader) throws GradingException {
+        if(context.phase() != Phase.Phase5) return;
+
+        try {
+            Set<File> testFiles = reader.filesMatching(".*client/src/test/java/.*").collect(Collectors.toSet());
+            boolean hardCodedPortFound = false;
+            boolean dynamicPortLineMissing = true;
+            for(File file : testFiles) {
+                for(String line : reader.getFileContents(file)) {
+                    if (line.contains(String.valueOf(CHECK_HARDCODED_PORT))) {
+                        hardCodedPortFound = true;
+                    }
+                    if(line.contains(DYNAMIC_PORT_LINE)) {
+                        dynamicPortLineMissing = false;
+                    }
+                }
+            }
+
+            if(hardCodedPortFound) {
+                context.observer().notifyWarning(String.format("Found hardcoded number (probably port) %d in client " +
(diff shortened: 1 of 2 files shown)


Fiwafoofa (approved):

src/main/java/edu/byu/cs/autograder/compile/verifers/ServerFacadeTestPortVerifier.java
```diff
@@ -0,0 +1,53 @@
...
+    public void verify(GradingContext context, StudentCodeReader reader) throws GradingException {
+        if(context.phase() != Phase.Phase5) return;
+
+        try {
+            Set<File> testFiles = reader.filesMatching(".*client/src/test/java/.*").collect(Collectors.toSet());
+            boolean hardCodedPortFound = false;
+            boolean dynamicPortLineMissing = true;
+            for(File file : testFiles) {
+                for(String line : reader.getFileContents(file)) {
+                    if (line.contains(String.valueOf(CHECK_HARDCODED_PORT))) {
```
frozenfrank, line 31: We should be doing a much more sophisticated regex comparison here.

src/main/java/edu/byu/cs/autograder/compile/verifers/ServerFacadeTestPortVerifier.java
```diff
@@ -0,0 +1,53 @@
...
+import edu.byu.cs.autograder.compile.StudentCodeVerifier;
+import edu.byu.cs.model.Phase;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ServerFacadeTestPortVerifier implements StudentCodeVerifier {
+    private static final String DYNAMIC_PORT_LINE = "port = server.run(0)";
```
frozenfrank, line 15: I know that this has already been merged, but I'm concerned that this check is too specific. Could a student feasibly write the line differently, but still accomplish the same purpose?

## Overview
Students could _very easily_ write slightly different versions of the line we expect them to write, and those variations will cause them to fail this check, even if their code would work in production.

Here are a few possible examples:
| Alternative Code Snipped | Variation |
|--------|--------|
| `port = server.run(0)` | None. Only accepted solution right now |
| `port =  server.run(0)` | extra space |
| `port_for_server = server.run(0)` | different variable name |
| `p = server.run(0)` | abbreviated variable name |
| `port = myServer.run(0)` | different class name |
| `port = server.startWithPort(0)` | different method name |
| `port = server.startWithAutoPort()` | different call signature |

## Potential Resolutions
There are basically two ways we can address the potential overflagging (false positives) of issues produced by this verification:
1. Mitigate the consequences of being flagged.
    - Ensure that this appears only as a warning, and allows the student to still receive a grade if it turns out that the code works anyways
2. Generalize the check to account for a wider variety of code snippets that still produce the same effect
    - Use a regex to allow for different kinds of whitespace, variable names, and method names etc

It will be tricky to define exactly when their code will work correctly based on _only_ static analysis of their code. As shown in the examples above, there are diverse ways to write the code that would all achieve the same function, but would be very tricky to define via regex with 100% accuracy.

Defining the pattern would be more possible if we knew/could assume/could define some characteristics of the way this behavior _must_ be implemented. This would effectively strike out some of the example possibilities in the table above and define a computable definition for compliance. While suggestions could happen at the starter code level— such as providing a class name, method name, and call signature—, the students would still be free to change that code to fail our pattern matching. The best approach would require a specific method name and call signature, and we would need to test it and clearly request it; this would give us a very reliable point of reference.

Therefore, I suggest that we go towards approach number 1. While we may already been generating a mere warning, we should also use language that clearly communicates our possibility for error. Some people who receive the warning may not need to actually pay any attention to it.

## Therefore, What?
If this argument is sufficiently valid, then I'll create a new issue to track an improvement/extension of this verification. We can continue discussing the best approach to take in the new issue, and link new PR's to it.

19mdavenport: Autograder warnings are similar to code warnings. Potential Resolution 1 was never not the case. We're checking for things we think will prevent them from getting full points, but still continuing with autograding in case it doesn't. 

Also, this line is a line in the starter code they shouldn't change, and if they do, they either change the type, which is allowed by this `var port = server.run(0);` to `int port = server.run(0);`, which wouldn't be flagged here, or they change the port `var port = server.run(8080);` which will probably cause the tests to fail (if they hardcode 8080, other hardcoded ports may work)

frozenfrank: Touché. Thanks for clarifying.

PR #394, opened 2024-06-25
382 backend record instances of previous tests failing in submission table

19mdavenport:
This adds functionality to record instances of grading issues (compilation issues, previous tests failing) in the submission table so that the students can see that output after closing the tab and the TA's can see that information.
This also fixes a couple of bugs/oversights that were either bothering me or I found while creating this.

@Fiwafoofa I'm not entirely certain of some of the wording. Could you test it out and see if you have any better ideas for some of the messages?

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/Grader.java
+++ b/src/main/java/edu/byu/cs/autograder/Grader.java
@@ -80,10 +80,11 @@ public class Grader implements Runnable {
 
     public void run() {
         observer.notifyStarted();
+        CommitVerificationResult commitVerificationResult = null;
         try {
             // FIXME: remove this sleep. currently the grader is too quick for the client to keep up
             Thread.sleep(1000);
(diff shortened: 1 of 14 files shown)


Fiwafoofa (approved):
I tested it and it worked for an already working submission and submission with a previous phase's test failing. As for the message, the only improvements I can think if is potentially including which phases' tests are failing with the message in `PreviousPhasePassoffTestGrader`; however, if a student just looks at the details, that information is already there, so it may not be worth adding. Another minor thing is a bit of redundancy with error message `Failed previous tests. Cannot pass off until previous tests pass`. It appears in like four different spots, twice in the student's info panel and twice in the submission panel as the `result notes` and `notes`. But personally, I don't mind it that much as it makes very clear what went wrong. Maybe you could modify the message to be `Failed previous phases' tests. Cannot pass off until previous tests pass.` ? Personally, I find the error messages fine. Is there something that doesn't meet expectation with the existing messages?

PR #397, opened 2024-07-01
Extra imports working

Fiwafoofa:
Add a verifier to read the maven compilation error output and check if there are any bad imports. If so, notify the observer of the bad imports on their result notes.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
@@ -12,6 +12,7 @@ import edu.byu.cs.util.ProcessUtils;
 import java.io.IOException;
 import java.util.Collection;
 import java.util.List;
+import java.util.Set;
 
 public class CompileHelper {
     private final GradingContext gradingContext;
@@ -66,7 +67,18 @@ public class CompileHelper {
         try {
             ProcessUtils.ProcessOutput output = ProcessUtils.runProcess(processBuilder, 90000); //90 seconds
             if (output.statusCode() != 0) {
-                Rubric.Results results = Rubric.Results.textError("Your Java source code could not be compiled", getMavenError(output.stdOut()));
+                String mavenError = getMavenError(output.stdOut());
+                String notes = "Your Java source code could not be compiled. See Details.";
+
+                Set<String> extraImports = NoExtraImportsVerifier.findExtraImports(mavenError);
+                if (!extraImports.isEmpty()) {
+                    notes += " It seems that you have imports from packages you cannot use. " +
+                            "Double check the Phase specs for the proper packages to import. " +
+                            "Please remove the following packages/imports from your code: " +
+                            String.join(", ", extraImports) + "." ;
+                }
+
+                Rubric.Results results = Rubric.Results.textError(notes, mavenError);
                 throw new GradingException("Failed to compile", results);
             }
         } catch (ProcessUtils.ProcessException ex) {
--- a/src/main/java/edu/byu/cs/autograder/compile/verifers/NoExtraImportsVerifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/verifers/NoExtraImportsVerifier.java
@@ -1,0 +1,19 @@
+package edu.byu.cs.autograder.compile.verifers;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class NoExtraImportsVerifier {
+
+    public static Set<String> findExtraImports(String mavenOutput) {
+        Set<String> extraImports = new HashSet<>();
+        Pattern pattern = Pattern.compile("package (.+) does not exist");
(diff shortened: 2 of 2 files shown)


src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
```diff
@@ -66,7 +67,18 @@ private void packageRepo() throws GradingException {
...
-                Rubric.Results results = Rubric.Results.textError("Your Java source code could not be compiled", getMavenError(output.stdOut()));
+                String mavenError = getMavenError(output.stdOut());
+                String notes = "Your Java source code could not be compiled. See Details.";
+
+                Set<String> extraImports = NoExtraImportsVerifier.findExtraImports(mavenError);
+                if (!extraImports.isEmpty()) {
+                    notes += " It seems that you have imports from packages you cannot use. " +
+                            "Double check the Phase specs for the proper packages to import. " +
+                            "Please remove the following packages/imports from your code: " +
+                            String.join(", ", extraImports) + "." ;
```
19mdavenport, line 78: This does not contain a mention of where specifically they imported something improperly. A student who mistakenly imported server classes in their client could possibly read this and think they cannot use anything from their server anywhere (including elsewhere in the server, or their unit tests, both of which are allowed). Do you think this could output where the bad import is from, or perhaps ignore bad imports from packages that do exist in their code?

src/main/java/edu/byu/cs/autograder/compile/verifers/NoExtraImportsVerifier.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class NoExtraImportsVerifier {
+
+    public static Set<String> findExtraImports(String mavenOutput) {
+        Set<String> extraImports = new HashSet<>();
+        Pattern pattern = Pattern.compile("package (.+) does not exist");
```
19mdavenport, line 12: Could this be a constant so it doesn't have to be reallocated every time?

src/main/java/edu/byu/cs/autograder/compile/verifers/NoExtraImportsVerifier.java
19mdavenport, line 1: Does this need to be it's own class? Could this be a helper method inside CompileHelper?

19mdavenport (changes requested):
I'm not completely sure this is needed anymore. I believe the original intent was to catch people attempting to use external libraries to their code. For the packages that do exist in another module, we already have a check for that. Do you think this code is necessary still? If so, could we make it so it doesn't duplicate what the module independence checker already says?

src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
Fiwafoofa: Good catch. I forgot about your module independence verifier and currently there is duplication. 

> This does not contain a mention of where specifically they imported something improperly.

The mention of where something was imported improperly is given in the actual Maven output in the text results and I omitted that as I did not want to put too much text inside that little notes box and it would have more or less be copy and pasted from the actual Maven error which students can see if they just click the details button.

> A student who mistakenly imported server classes in their client could possibly read this and think they cannot use anything from their server anywhere (including elsewhere in the server, or their unit tests, both of which are allowed).

I can reword the message and rework the check to emphasize that this is meant to be talking about external libraries and not the modules inside the project.

> Do you think this could output where the bad import is from, or _perhaps ignore bad imports from packages that do exist in their code?_

I'm not sure what you mean by bad imports and ignoring them. Can you clarify?

Fiwafoofa: Oh, I think by bad imports you mean cases where the client imports packages from the server or vice versa.

Fiwafoofa:
After more thinking, I agree that this may not be needed anymore due to the work it requires and the existence of the module independence verifier. If students really just look at the details, I feel they can figure out a lot on their own without the autograder explicitly telling them. But this did get me thinking, this issue wouldn't be a problem if people strictly stuck to the information given in the `pom.xml`s. Maybe I'm misunderstanding how the build system works in IntelliJ, but is there a reason we sort of use two different pieces of software to manage this projects dependencies (Intellij's proprietary software and Maven's)? There ends up being a discontinuity in projects because of that. If we just strictly chose to use Maven for dependency management, I feel like we wouldn't have this issue of people doing bad imports as the IDE would follow what Maven says, but I really don't know if I'm right due to my lack of knowledge of how IntelliJ works. Michael, maybe you know more about the technical details?

19mdavenport:
That does sound correct. However, Maven can be a bit temperamental sometimes, so the professors recommend using the IntelliJ system to students. The auto-grader can't use IntelliJ though, so it relies on maven, which is why we have both. This could also be solved by students not ever adding dependencies other than what the specs/getting started say (assuming the getting started is always correct, which has been variable, I'll check those again), but they always manage to add dependencies

PR #403, opened 2024-07-26
add check for checkstyle successfully finishing

### LINKED ISSUE #402: Backend: Crashing the code quality grader can give unearned points
As @cjvuoksi found, despite my best efforts, it is possible to crash the quality grader. If this happens, it prevents the quality grader from getting to all of the code and the auto-grader can award unearned points. It should be easy to detect crashes (checkstyle always says something like "audit done" or similar after completing, so without that line, we can know it crashed). I will try to fix crashes as I become aware of them, but if crashes happen, we should handle that in some way.

Personally planning on giving a zero and directing students to see a TA, but @softwareconstruction240/tas does anyone else have any other ideas of what to do in that situation?

19mdavenport:
Based off my previous PR and should be merged in after.

Resolves #402

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
+++ b/src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
@@ -47,6 +47,10 @@ public class QualityAnalyzer {
             throw new GradingException("Error running code quality: " + e.getMessage(), e);
         }
 
+        if(!checkstyleFinished(output)) {
+            return new QualityAnalysis(0, "", "Could not complete code quality analysis. Please go see a TA.");
+        }
+
         output = output.replaceAll(stageRepo.getAbsolutePath(), "");
         output = output.replaceAll(stageRepo.getPath(), "");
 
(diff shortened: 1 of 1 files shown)


src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
```diff
@@ -47,6 +47,10 @@ public QualityAnalysis runQualityChecks(File stageRepo) throws GradingException
             throw new GradingException("Error running code quality: " + e.getMessage(), e);
         }
 
+        if(!checkstyleFinished(output)) {
+            return new QualityAnalysis(0, "", "Could not complete code quality analysis. Please go see a TA.");
```
Fiwafoofa, line 51: I'm not too familiar with the code quality output, so you can feel free to disregard this, but if the checkstyle does crash and the crash outputs some useful information, is it possible to output that in the results?

src/main/java/edu/byu/cs/autograder/quality/QualityAnalyzer.java
```diff
@@ -163,6 +167,10 @@ private boolean categoryPasses(QualityRubricCategory category, QualityOutput out
         return true;
     }
 
+    private boolean checkstyleFinished(String output) {
```
Fiwafoofa, line 170: Top tier code quality with the method here :)

Fiwafoofa (approved):

PR #407, opened 2024-08-08
fix late penalty scores

19mdavenport:
Fixed by changing the `applyLatePenalty` method to return a new Rubric rather than alter an existing rubric, then determining score and sending to canvas with the rubric with the penalty applied but generating the submission object with the rubric without the penalty.

Like I said in the issue, can be confusing: 
image (auto-grader rubric)
image (canvas rubric of same submission)
but hopefully a less confusing and/or a better type of confusing for students

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/score/Scorer.java
+++ b/src/main/java/edu/byu/cs/autograder/score/Scorer.java
@@ -63,8 +63,8 @@ public class Scorer {
         }
 
         int daysLate = new LateDayCalculator().calculateLateDays(gradingContext.phase(), gradingContext.netId());
-        applyLatePenalty(rubric.items(), daysLate);
-        float thisScore = getScore(rubric);
+        Rubric lateAppliedRubric = applyLatePenalty(rubric, daysLate);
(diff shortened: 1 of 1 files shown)


src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -110,17 +110,16 @@ private Rubric transformRubric(Rubric rubric) throws GradingException, DataAcces
      * @throws DataAccessException When the database can't be reached.
      * @throws GradingException    When other conditions fail.
      */
-    private Submission successfullyProcessSubmission(
-            Rubric rubric, CommitVerificationResult commitVerificationResult,
-            int daysLate, float thisScore
-    ) throws DataAccessException, GradingException {
+    private Submission successfullyProcessSubmission(Rubric rubric, Rubric lateAppliedRubric,
```
Fiwafoofa, line 113: You might want to update the Javadoc for the new rubric parameter.

Fiwafoofa (approved):
👍

Issue #408, opened 2024-09-05
Fullstack: Commit Approval buttons don't appear on submissions to the GitHub phase


19mdavenport:
Also, the score for the github phase should probably always be full points if it can find the repository

19mdavenport:
On suggestion from @leesjensen, we could also fix this by just allowing students to resubmit this one assignment.

PR #411, opened 2024-09-05
410 backend remove git commit requirements for GitHub phase

DallinFromEarth:
This in theory should fix the issues mentioned in the TA meeting today, where students were getting flagged with not enough commits on the GitHub phase, and the admin approval buttons weren't showing up for the GitHub phase.

I'm having issues getting the autograder authorized with canvas on my own machine, so I haven't been able to test this yet. I would appreciate someone running this locally on their machine when then can.


19mdavenport:
Also resolves #408 since it will no longer be relevant

19mdavenport (approved):
Sorry I had to do some administrative changes but I'd like to deploy this as soon as possible

Issue #412, opened 2024-09-06
Backend: URLs not matching regex bug


19mdavenport:
URL's not matching the regex can cause submissions to hang in the queue. If a GradingException is thrown from the Grader contructor, weird things happen and the netid is left in the grading queue.

19mdavenport:
Also, putting in www is technically still a valid repository link even if it's not necessary, so may need to add that to the regex. Also also, when the github link changes, the user object isn't actually refreshed.

frozenfrank:
> Also also, when the github link changes, the user object isn't actually refreshed.
@19mdavenport I'm not sure what is being referred to with this statement. Is this an update that needs to happen in the front-end code, or back-end related? Where in the code does the `user` object need to be refreshed?

frozenfrank:
> So for the first submission after changing links in canvas, it grades the old repo link and then every submission after is the new link.

@19mdavenport Thanks for clarifying that behavior. This information has been extracted into it's own issue #418.

frozenfrank:
@19mdavenport Is this issue entirely covered by #417 and #418? If so, I think this is now complete, but I want to confirm with you before closing it.

TheDavSmasher:
@19mdavenport ?

TheDavSmasher:
I believe this issue is complete unless @19mdavenport or @frozenfrank have something to say about it, in which case it will be re-opened

Issue #409, opened 2024-09-05
Fullstack: Accept student GitHub URL's from AutoGrader


frozenfrank:
## Issue
Currently, for the initial "Chess GitHub Repository" assignment, students must submit information in two different places:
1. Submit GitHub URL to Canvas (to identify the student's repo)
2. Click the button on the AutoGrader to evaluate it and receive points

Currently, this requirement is documented in the ["Deliverables"](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/chess-github-repository/chess-github-repository.md#-deliverable) section of the assignment spec.

We've noticed that in the Fall 2024 semester, a significant portion of students (as many as 16 of 51; 31%) have been submitting the url to Canvas but not requesting the grade from the autograder. This leads to TA's sending individual messages to the students who haven't fully submitted, and student's feeling confused when they think they've finished the assignment, but it's not yet fully done.

Is there a better way to do it?

## Proposal
Adjust the autograder to accept and require the GitHub URL from students before handling autograding requests.

### GitHub URL Acceptance Specifics
The GitHub URL should be sent to Canvas for now to remain compatible with our current system. However, in the future, we may choose to keep it fully internal to the AutoGrader and only send grades to Canvas. 

We should provide basic validation that the URL is a GITHUB url before accepting it into our system. 

This would also be a good opportunity to perform a repository claim check so that a student cannot claim a repo already in use (this would be an improvement upon #69).

**Consider: Do we want to allow students to edit their GitHub URL?** This would be helpful if students accidentally provide an incorrect URL, but this can be minimized with URL validation and repository claim validation. However, we may not want to allow that kind of behavior since it could lead to increased cheating. Firstly, the commit verification system may always produce failing results if there is no shared history with a previous submission. Secondly, this may only provide students the opportunity to claim work they didn't set out to claim as theirs in the beginning of the semester.

### Potential Front End Implementations

Below are two ideas for ways this could be implemented. Please comment with your vote and provide additional insight or alternative implementation ideas. We could move forward with any one, or a combination of these ideas mixed with new ideas.

#### Proposal 1: Onboarding Flow
Craft an introductory "onboarding flow" into the autograder that requires a GitHub URL from the student before proceeding to the main page. This page would block students from seeing the rest of the information in the autograder until they have a GitHub URL saved. In the future, we may add additional steps in the flow to provide helpful information or to ask additional questions.

##### Pros
* Obviously asserts the information we need
[...]

PR #417, opened 2024-09-11
Extend regex to accept www in the URL

frozenfrank:
This addresses the following specific portion of #412:
> Also, putting in www is technically still a valid repository link even if it's not necessary, so may need to add that to the regex.

Test cases have also been added to verify that the autograder continues to acknowledge the new cases.


19mdavenport:
Looks good so far. I don't see new test cases. Have you already made those or are just planning to?

frozenfrank:
@DallinFromEarth Also pointed out that we should standardize all URLs into the same format. Currently, we preserve the `.git` if the student includes it, or keep it off if it's not provided. This will be accounted for in this PR.

DallinFromEarth:
> @DallinFromEarth Also pointed out that we should standardize all URLs into the same format. Currently, we preserve the `.git` if the student includes it, or keep it off if it's not provided. This will be accounted for in this PR.

Yes. For the sake of being able to easily compare repo's between each other, we should standardize the format in which the URLs are saved. Work in #409 needs to be able to easily search and compare repo links to ensure no one has the same repo as another student. (Likely searching just through a new SQL table). `https://github.com/yourmom/cheese` and `https://github.com/yourmom/cheese.git` would not be flagged as duplicates, despite pointing to the same repo. Instead of having to clean the urls every time we look at them, we should just clean and standardize them once when we save them.

Current language on the front end for #409 tells students to match this pattern: `https://github.com/{USERNAME}/{REPO}`

frozenfrank:
@19mdavenport @DallinFromEarth This is the [regex101](https://regex101.com/r/r3dDAW/6) we're using to compose and test the extended behavior. Notice that we have multiple test cases each on separate lines. Additional test cases have been added to cover "www." in the URL, and we've also made other adjustments. 

Can you please review the regex and associated test cases to make sure that we've covered everything? If not, please update the regex within Regex101 and/or comment about any observed shortcomings in this thread.

19mdavenport:
I guess depending on how thorough you're trying to be https://github.com:443/softwareconstruction240/autograder is valid but not covered

DallinFromEarth:
> https://github.com:443/softwareconstruction240/autograder 

For the love of all that is holy on God's green earth, don't accept that. If students are trying to make things more complicated than simply copying the link provided by clicking the "code" button on their repo, then that's their problem.

frozenfrank:
@19mdavenport @DallinFromEarth This PR is now ready for review and merging.

The [test cases](https://regex101.com/r/r3dDAW/9) have been expanded, and the test running code has been adjusted to more completely validate that the correct results are coming back. @DallinFromEarth Note that in the interest of accepting as many valid URLs as possible, the code has been adjusted to accept ports in the URL. Applying solid software design principles from CS 340, Design choices were also made to make the system as GitHub agnostic as possible (even though it is currently requiring GitHub urls).

I now have full confidence that this regex parsing works exactly when it should, and rejects when it should.
image

frozenfrank:
After thinking about this over the weekend, I've changed my mind and now agree with @DallinFromEarth: we should not support URLs with ports in them.

For standardization purposes, we're currently stripping off  the port anyways which changes the link. If a student submits a link with a port, we're going to remove it during back end processing (not obvious to the student). If the link would have only worked if the port was included, then a working link may be changed into a non-functional link in a non-obvious way.

Instead of creating this opportunity for potential confusion, we should simply reject the link and cause them to deal with the issue upfront. This may be as simple as removing the port, or it may require them to use a different link that doesn't require the port.

frozenfrank:
After discussing this with @19mdavenport, we decided to move the URL parsing code out of `Grader` into a utility class that could be used in any portion of the future development.

frozenfrank:
@19mdavenport @DallinFromEarth The refactor of the code into a separate class has been completed, and this PR is now ready to review and merge. Note that it also includes #436 which has been completed and will be reviewed as part of this PR.

I chose to extract the code into a class called `RepoUrlValidator` in the `utils/` package because it most closely represents its current function. This utility may be generalized or renamed or moved in the future, but based on the present conditions, this seems to be the best place for it.

The tests for the method were also moved.

Additionally, the code was moved to instance methods which more properly respect OO code, and `static` wrapper methods were added as a convenience for users of the code.

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
```diff
@@ -0,0 +1,106 @@
...
+        String repoName = "([\\w.-]+?)";
+        return new String[]{
+                "^(?:https?://)?(?:www\\.)?"+domainName+"/"+userName+"/"+repoName+"(?:\\.git|/|\\?|#|$)", // https
+                "^git@"+domainName+":"+userName+"/"+repoName+"\\.git$" // ssh
+        };
+    }
+
+    public static class InvalidRepoUrlException extends GradingException {
+        InvalidRepoUrlException(String message) {
+            super();
```
frozenfrank, line 102: This call to `super();` should pass in the parameter received in the constructor call. Robust linting should have warned about this unused variable.

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
```diff
@@ -0,0 +1,106 @@
...
+        String githubUsername;
+        String repositoryName;
+        for (String regexPattern : getRepoRegexPatterns()) {
+            pattern = Pattern.compile(regexPattern, Pattern.CASE_INSENSITIVE);
+            matcher = pattern.matcher(trimmedRepoUrl);
+            if (matcher.find()) {
+                domainName = matcher.group(1).toLowerCase();
+                githubUsername = matcher.group(2);
+                repositoryName = matcher.group(3);
+                return String.format("https://%s/%s/%s", domainName, githubUsername, repositoryName);
```
frozenfrank, line 53: A more generic solution would abstract this exact url formatter, and ideally it would associate the extracted with the formatter. That would keep those too closely related pieces of functionality coupled together as they should be.

frozenfrank (commented):
There are a few non-critical adjustments which could be made before this is merged.

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
frozenfrank: This fits our current use case just fine. If we need to change this, we'll need to modify the source code anyways. This is not following the open-closed principle, but we can accept that for now.

19mdavenport (approved):

PR #423, opened 2024-09-17
Head hash missing improvement

### LINKED ISSUE #420: Backend: AutoGrader rejects submissions without the previous head commit without grading
## Overview
Too many students are running into project setup issues, and nuking their existing Git repository just to replace it with another one. This seems fine, except that the AutoGrader assumes that their previous commit history is available in order to grade the new commits. (It could be available, but students end up deleting the previous history which really isn't best practice.)

I tried having the student resubmit the GitHub phase (to produce a reference to a valid Git object), but the AutoGrader doesn't seem to be using it. Or, if it is, it's still getting caught up by the missing object reference.

The result is a failure message reported to the student and a refusal to continue grading. This means that no manual submission entry exists in the database to approve.
> Failed to verify commits: Missing commit OLD_COMMIT_HASH
> — src/main/java/edu/byu/cs/autograder/git/GitHelper.java:150

## Complexity
The Commit Verifier is specifically expecting to find the previous commit hash in the repository so that we can grade only the new commits. 

## Proposal
In the event that some commits cannot be found, we should generate a warning on Phase 0, and on future phases, continue the grading but fail the verification so that a TA can approve the submission.

Reviewed with @19mdavenport 

## TODO
- [ ] Provide counts of the number of commits that would be affected in each warning message
- [ ] Determine if we should properly re-evaluate all of the remaining commits after some are removed (significant commits, days with commits)
- [ ] Send warnings to Observer with `sendWarning()`
- [ ] Test that the data structures are behaving the way we expect them to

frozenfrank:
This changes the behavior when errors arise during commit verification. With this change, grading will continue and ultimately report "unverified;" a TA will be able to approve the submission like with other issues.  It also provides additional documentation to prevent future misunderstandings in the future.

This specifically changes the behavior when a student requests grading for a new phase when their previous submission commit is missing. This was happening frequently as students recreated repos between the GitHub phase and Phase 0, typically due to folder issues.

Resolves #420.

### DIFF
--- a/src/main/java/edu/byu/cs/analytics/CommitAnalytics.java
+++ b/src/main/java/edu/byu/cs/analytics/CommitAnalytics.java
@@ -240,7 +240,7 @@ public class CommitAnalytics {
     private static CommitsBetweenBounds produceCommitIterable(Git git, ObjectId headObjId, ObjectId tailObjId)
             throws IncorrectObjectTypeException, MissingObjectException, GitAPIException {
         try {
-            return new CommitsBetweenBounds(git.log().addRange(headObjId, tailObjId).call(), false);
+            return new CommitsBetweenBounds(git.log().addRange(tailObjId, headObjId).call(), false);
         } catch (MissingObjectException missingObjectException) {
             return produceCommitIterable(git, headObjId, true);
         }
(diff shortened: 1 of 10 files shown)


frozenfrank:
I still need to add a new test to verify that this doesn't ever regress in the future.

frozenfrank:
@19mdavenport This initial version is now ready for review. Expect another PR with further improvements to this area of the code.

19mdavenport (approved):

PR #424, opened 2024-09-18
Update compose.yml use latest loki image

### LINKED ISSUE #421: Backend: Grafana logs should never clear
Today, while looking into an issue, we found that the Grafana logs had been cleared from the autograder. This removed necessary debugging information.

19mdavenport:
I still haven't fully tested to make sure this doesn't break anything else, but this should solve the specific problem with the logs
resolves #421

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -3,7 +3,7 @@ networks:
 
 services:
   db:
-    image: mysql:latest
+    image: mysql:9.0.1
     profiles:
       - with-db
     restart: always
@@ -36,7 +36,7 @@ services:
       - autograder
 
   loki:
-    image: grafana/loki:2.9.0
+    image: grafana/loki:2.9.10
     ports:
       - "3100:3100"
     command: -config.file=/etc/loki/local-config.yaml
@@ -67,7 +67,7 @@ services:
           editable: false
         EOF
         /run.sh
-    image: grafana/grafana:latest
+    image: grafana/grafana:11.2.0
     ports:
       - "3000:3000"
     networks:


compose.yml
```diff
@@ -3,7 +3,7 @@ networks:
 
 services:
   db:
-    image: mysql:latest
+    image: mysql:9.0.1
```
pawlh, line 6: version locking is great idea.

Not necessary by any means, but most images allow you the granular control to lock on to specific major, minor, or patch versions. Instead of locking onto `mysql:9.0.1`, you could also do `mysql:9` or `myql:9.0` which would keep you up to date on the latest minor and patch versions respectively without risking jumping major versions

pawlh: https://hub.docker.com/_/mysql/tags

pawlh: Although it looks like it is not an option for `grafan/loki` or `grafana/grafana` 😢

frozenfrank (commented):
The changes look reasonable to me.

In this case, we were facing an issue because we had fixed to an older version that didn't include big fixes. Any reason we wouldn't want to be using the latest versions?

frozenfrank (approved):

pawlh:
> Any reason we wouldn't want to be using the latest versions?

A good reason is for stuff like [this](https://github.com/blakeblackshear/frigate/releases/tag/v0.14.0) or [this](https://dev.mysql.com/blog-archive/upgrading-to-mysql-8-0-here-is-what-you-need-to-know/) where the major update might require the user to manually make changes that can't be automated.

`:latest` can break your deployment in environments like the autograder where containers automatically update themselves. Version locking the image is the closest docker has to things like [`package-lock.json`](https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json) or [`go.sum`](https://go.dev/ref/mod#go-sum-files)

Although to be fair this is all mostly theoretical. It seems unlikely that this project relies on any features of `mysql`, `grafana`, or `loki` that will be significantly changed anytime soon

PR #425, opened 2024-09-18
Manage Repo URLs in the Autograder

DallinFromEarth:
### Overview
This pull request moves the management of repo URLs from canvas, and makes it completely managed by the Autograder itself.

Key changes include:
- Canvas is never queried for a student's repo URL
- Students are required to enter a repo URL the first time they log into the Autograder
- Students may change their repo URL anytime within the Autograder
- Admins may change a student's repo URL anytime from the Student Info window
- Changes in repo URLs are tracked, and available for admin inspection on a per-student basis
- Students are blocked from claiming repo URLs claimed by any other student, whether or not it is currently claimed by that student

#### What counts as a valid repo URL
The front end will not allow students to submit anything as a repo URL that doesn't start with `http` and contains `github.com/`. Its a very surface level check. The backend will then remove (if present) the `www.` from the beginning and `.git` from the end, to ensure consistent formatting. The backend will then attempt to clone the repo. If it can clone the repo, it is considered a valid URL and saved. If it can not, then it is rejected and the user is told to check their link is valid.

#### Changes to user registration
Previously, to register, a student had to submit the GitHub Repo Assignment on Canvas. Now, all they have to do is log-in to the Autograder. The Autograder will verify that they are a student in canvas, then create a new student record. Now that the student record exists in the Autograder, they will be prompted to submit a repo URL. If they log out and try to log-in without doing so, they still will be prompted to submit a repo URL.

#### Checking students don't claim another student's repo URL
When a user submits a request to change a repo URL, the new URL is checked against a new table in the database called `repo_update` containing a list of every time a user's repo URL has changed. If ANY user (besides the user whose repo URL is getting changed) has ever claimed that URL, the request to change is rejected. If a student submitted the request, they are just told to go talk to the TAs. If an admin submitted the request, they will see who has previously claimed that URL and when.

#### New backend endpoints
- `{PATCH} /api/repo` allows a student to change their repo URL. Pulls their netId from the session, and the new repo URL is passed in the request body
- `{PATCH} /api/admin/repo/:netId` allows an admin to change a student's repo URL. The student's netId is passed in via the request URL, and the new repo URL is passed in the request body
- `{GET} /api/admin/repo/history` allows an admin to get info from the new `repo_update` table. Allows table to be searched by `netId` and `repoUrl`, one or both. Parameters are passed in by a query string in the request URL

### Potential Future Work
While this feature is fully built out, there are some things that the team could reasonably want added or changed. Here are some things that came up during development that I figured should be decided later and done in later pull requests)
- While admins can view how often a student has changed their repo URL, they only can see successful changes. If they are blocked from changing their repo URL because it is/was claimed by another student, that attempt is not logged anywhere. It may or may not be beneficial to log those to see who is attempting to claim another student's repo URL. 
- Admins have no way of overriding a block. I can't see a scenario where a student should be allowed to claim the same repo URL as another student, but stranger things have happened. One could argue for the need for admins to be able to force change a student's repo URL, despite plagiarism checks.
- Currently, the page is forced to reload when an admin changes a student repo URL. To fix that, we'd need to store some data at a higher level than is currently stored.
- Run the GitHub phase at registration. This would reduce the steps a student has to take. But it could introduce additional complications depending on how the professors want things to work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,4 +41,3 @@ build/
 /logs/
 /phases/
 /commit-cache/
-/validation_cloning/
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ build/
 /logs/
 /phases/
 /commit-cache/
+/validation_cloning/
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -66,7 +66,7 @@ public class GitHelper {
                     skipCommitVerification(true, headHash, null);
         } catch (GradingException e) {
             // Grading can continue, we'll just alert them of the error.
-            String errorStr = "Internally failed to evaluate commit history: " + e.getMessage();
+            String errorStr = "Internally failed to evaluate commit history: " + e.getMessage() + " (this could happen because the repo url has been changed since the last phase, or the student did something to their commit history)";
             gradingContext.observer().update(errorStr);
             LOGGER.error("Failed to evaluate commit history", e);
             return skipCommitVerification(false, headHash, errorStr);
@@ -146,7 +146,7 @@ public class GitHelper {
             }
         } catch (IOException | GitAPIException | DataAccessException e) {
             var observer = gradingContext.observer();
-            observer.notifyError("Failed to verify commits: " + e.getMessage());
+            observer.notifyWarning("Failed to verify commits: " + e.getMessage());
             LOGGER.error("Failed to verify commits", e);
             throw new GradingException("Failed to verify commits: " + e.getMessage());
         }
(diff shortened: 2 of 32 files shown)


DallinFromEarth:
I just added a function in `UserController.java` called `loadCurrentReposIntoUpdateTable()`. This code is a temporary solution to a temporary problem. Since this new system is being added after all the students have already put their repo urls into the system, they haven't been put in the new `repo_update`. `loadCurrentReposIntoUpdateTable()` will be called anytime an admin updates the repo url for the test student. What it does, is it goes through the all the users, and if they don't have an update entry in the `repo_update` table, it will create one. This will let the claim-checker check new repo submissions against those repo urls. 

If anyone has a better idea of how to do this besides having this temporary code here, I'm all ears.

19mdavenport (approved):

PR #429, opened 2024-09-24
Spot fix: late penalties on manually approved submissions

19mdavenport:
This is a spot fix for the problem with late penalties on manually approved submissions. I discussed a more permanent solution with @frozenfrank earlier, but this is a spot fix to hold us over as the more permanent solution will take some time and this is fairly urgent.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/score/Scorer.java
+++ b/src/main/java/edu/byu/cs/autograder/score/Scorer.java
@@ -18,6 +18,7 @@ import edu.byu.cs.util.PhaseUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.EnumMap;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -130,18 +131,21 @@ private Submission successfullyProcessSubmission(Rubric rubric, Rubric lateAppli
...
-     *               Any items not set will be populated with their value from Canvas.
-     * @param phase The phase being graded
-     * @param netId Net ID of student being
+     * @param submission Submission object of approved submission
      * @param penaltyPct The approved GIT_COMMITS penalty percentage
      * @throws GradingException When preconditions are not met.
      * @throws DataAccessException When the database cannot be reached.
      */
-    public static void attemptSendToCanvas(Rubric rubric, Phase phase, String netId, int penaltyPct, String commitPenaltyMsg) throws GradingException, DataAccessException {
+    public static void attemptSendApprovedScoreToCanvas(Submission submission, int penaltyPct, String commitPenaltyMsg) throws GradingException, DataAccessException {
```
frozenfrank, line 139: I don't support this change in the call signature of this method.

The previous call signature has been intentionally destructured to require only the lowest possible  set of objects, and data. The method is clearly documented as "sending a provided rubric to Canvas," and it's already doing exactly that job.

Passing in more complicated data (`Submission`) and adding more dependencies (`new LateDayCalculator()`) is contradictory to solid software architecture and is violating several important principles. As already documented in the call signature, any/all penalties (late, commit verification, or otherwise) should already be applied before calling this method.

It's sad to see perceived urgency leading to unnecessary technical complexity and technical debt.

frozenfrank: Case in point, did you mean to also adjust the other call signature that used to have this name? It is also submitting Rubrics to Canvas, but it doesn't have this changed behavior. What new issues will that cause? Are there other use-cases that will still be left open because they are leveraging a different call signature?

How do we know that the late-penalty will not be applied twice for students submitting regularly (Once by the main system, and a second time by this system)?

src/main/java/edu/byu/cs/util/SubmissionUtils.java
```diff
@@ -75,8 +75,7 @@ public static void approveSubmission(
         // Send score to Grade-book
         float approvedScore = Scorer.prepareModifiedScore(withheldSubmission.score(), penaltyPct);
         String gitCommitsComment = "Submission initially blocked due to low commits. Submission approved by admin " + approverNetId;
-        Scorer.attemptSendToCanvas(withheldSubmission.rubric(), withheldSubmission.phase(), withheldSubmission.netId(),
-                penaltyPct, gitCommitsComment);
+        Scorer.attemptSendApprovedScoreToCanvas(withheldSubmission, penaltyPct, gitCommitsComment);
```
frozenfrank, line 78: Acknowledging the current model with `Submission`s having original scores without late penalties applied, this now becomes the correct place to apply the late penalties. The Scorer methods most reasonably assume that all calculations have already occurred. Doing it lower exposes the option of double-penalizing, besides the previously mentioned issues with dependencies.

frozenfrank (commented):
There is always a spectrum between _getting something done_ and _doing it right_. I personally feel like this is too far on the getting-it-done-quickly side, but I see that it does solve the immediate problem. 

Pushing it out may lead to students correctly being penalized for being late, but it's coming at the cost of additional technical debt. Proceed as you wish.

frozenfrank:
Now that #430 has been created to account for this, we can punt the issue down the road and push this out and fix the issue 🐞

19mdavenport:
@frozenfrank I intentionally did this in one commit so that I could revert that commit at the beginning of the long-term solution, which would solve the technical debt, would it not?

PR #431, opened 2024-09-25
add customization options to banner message

DallinFromEarth:
you can now put a link into the banner message that is clickable and change the color


19mdavenport:
Did you decide against making it expire?

DallinFromEarth:
> Did you decide against making it expire?

Good point. I did decide against it, but then I took it as a challenge. now it expires.

I doesn't disappear from a user's front end, but if they refresh or load the page after it expires, it wont be there anymore. This happens because now, everytime a user asks the server for the banner, it checks to see if its past expiration. if it is, it clears the banner from the database.

src/main/java/edu/byu/cs/controller/ConfigController.java
```diff
@@ -124,18 +179,30 @@ private static JsonObject getPrivateConfig() throws DataAccessException {
...
 
+    private static Instant getInstantFromUnzonedTime(String timestampString) throws DateTimeParseException {
+        ZoneId utahZone = ZoneId.of("America/Denver");
+
+        DateTimeFormatter formatter = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
+        LocalDateTime localDateTime = LocalDateTime.parse(timestampString, formatter);
+        ZonedDateTime utahTime = localDateTime.atZone(utahZone);
+
+        return utahTime.toInstant();
+    }
```
19mdavenport, line 204: Do we want to be hardcoding this? @frozenfrank what do you think?

frozenfrank:
It would be much better to read in the "America/Denver" string from a
configuration variable or slot in the database. This allows us to configure
it remotely, or deploy this more broadly with minimal code changes.

DallinFromEarth:
Is this something that we'll want to change across the app? Everything else in the autograder already assumes Utah time, at least as far as I'm aware

frozenfrank:
@DallinFromEarth @19mdavenport This configuration improvement is really part of #156.

While we have tried to minimize the number of times we hardcode the timezone, there currently appear to be **5 different** occurrences of the timezone definition in the code. Several of them already have comments indicating that we should extract the exact value into a config table.

image

I don't this we should hold up this PR on the config table improvements. We will already need to extract the constant from multiple other files in order to fully fulfill the request, and that isn't something we need to do now.

However, that doesn't mean we can just _assume_ that autograder is operating in the Utah timezone. We should practice technical diligence and avoid making decisions for as long as possible (good software design principles).

src/main/java/edu/byu/cs/controller/ConfigController.java
frozenfrank: Let's save this improvement in #156.

19mdavenport (approved):

PR #433, opened 2024-09-26
416 canvas preserve comprehensive startup nature of GitHub assignment

19mdavenport:
Re-adds the commit verification into the GitHub assignment grader, with caveats:

- Minimum line count for commits is now zero (as well as minimum number of days, although this shouldn't matter)
- Instead of giving zeros and making the student talk to a TA, the submission "fails" (like a submission not passing all passoff tests for phase 0) and they can resubmit without talking to a TA


DallinFromEarth (approved):
Looks great to me. Ran the code on my end and it all seemed to work as prescribed

src/main/java/edu/byu/cs/autograder/Grader.java
```diff
@@ -124,7 +124,7 @@ private Rubric evaluateProject(RubricConfig rubricConfig) throws GradingExceptio
                     case PASSOFF_TESTS -> new PassoffTestGrader(gradingContext).runTests();
                     case UNIT_TESTS -> new UnitTestGrader(gradingContext).runTests();
                     case QUALITY -> new QualityGrader(gradingContext).runQualityChecks();
-                    case GITHUB_REPO -> new GitHubAssignmentGrader().grade();
+                    case GITHUB_REPO -> new GitHubAssignmentGrader().grade(commitVerificationResult);
```
frozenfrank, line 127: I like this division of responsibility where the "Phase Grader" is now responsible for verifying the state of the commit verification result.

src/main/java/edu/byu/cs/autograder/git/GitHubAssignmentGrader.java
```diff
@@ -11,11 +11,20 @@
  * then this grader will give them full points.
  */
 public class GitHubAssignmentGrader {
+    private static final String RESUBMIT_PROMPT = " Push another commit to your repository and re-request grading to receive points for this assignment.";
```
frozenfrank, line 14: Good use of configurable string messages. I'm glad to see that they are giving clear instructions to the students to resubmit for credit.

src/main/java/edu/byu/cs/autograder/git/GitHubAssignmentGrader.java
```diff
@@ -11,11 +11,20 @@
...
 
-    public Rubric.Results grade() throws DataAccessException {
+    public Rubric.Results grade(CommitVerificationResult commitVerificationResult) throws DataAccessException {
         RubricConfig rubricConfig = DaoService.getRubricConfigDao().getRubricConfig(Phase.GitHub);
         RubricConfig.RubricConfigItem configItem = rubricConfig.items().get(Rubric.RubricType.GITHUB_REPO);
-        return new Rubric.Results("Successfully fetched your repository", 1f, configItem.points(), null, null);
-    }
 
+        if(commitVerificationResult.verified()) {
+            return new Rubric.Results("Successfully fetched your repository and verified commits",
```
frozenfrank, line 22: It probably makes sense to extract this string into a `SUCCESS_MESSAGE` variable for consistency.

src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -280,14 +281,15 @@ private static CanvasRubricAssessment constructCanvasRubricAssessment(Rubric rub
     }
 
     private boolean passed(Rubric rubric) {
-        if(!PhaseUtils.isPassoffRequired(gradingContext.phase())) {
-            return true;
+        Collection<Rubric.RubricType> requiredTypes = PhaseUtils.requiredRubricTypes(gradingContext.phase());
```
frozenfrank, line 284: This code looks a lot more general and reusable. Nice work!

frozenfrank (approved):
This all look good to me. I think we should extract out the `SUCESS_MESSAGE` prompt, but otherwise, it looks fantastic.

## Regrading Commit History for GitHub phase
Do we want to regrade the commit history each time for this phase? Previous discussions on this topic acknowledged that it doesn't happen, but it would actually be a small change to allow it to happen. 

Specifically, the `GitHelper:verifyCommitRequirements()` method is responsible for returning the verification result. It explicitly looks for an existing result and returns that cached version; otherwise, it initializes the full grading process. This would be the correct place to provide a `PhaseUtils.allowCachedCommitVerificationResults()` which would enable the caching behavior by default while allowing us to turn it off for this phase.

src/main/java/edu/byu/cs/autograder/Grader.java
19mdavenport: Thanks. That was @Fiwafoofa 's idea originally, I'm just readding his code

frozenfrank:
@19mdavenport Are we waiting on anything else before merging this PR?

PR #435, opened 2024-09-26
430 fullstack enhance late penalty applied presentation experience

19mdavenport:
I was surprised how little I actually had to change to do this (I guess that means we're generally writing ok code).

This adds a "raw score" to both rubric items and submissions and displays it to the user so that students are less confused about why they are losing points (actually losing points or late penalties)

@DallinFromEarth @frozenfrank I'd be interested in any thoughts you guys had on how to make this look better. I tried fairly hard to make it look as good as I could, but you might have some ideas

This does include editing the database, you can update your tables by doing the following:
```
ALTER TABLE `submission` ADD COLUMN `raw_score` FLOAT NOT NULL AFTER `score`; 
UPDATE `submission` SET `raw_score` = `score` WHERE `raw_score` = 0;
```


src/main/java/edu/byu/cs/model/Rubric.java
```diff
@@ -55,10 +40,15 @@ public record RubricItem(
     public record Results(
             String notes,
             Float score,
+            Float rawScore,
             Integer possiblePoints,
             TestAnalysis testResults,
             String textResults
     ) {
+        public Results(String notes, Float score, Integer possiblePoints, TestAnalysis testResults, String textResults) {
```
frozenfrank, line 48: I love this constructor overload that maximizes backwards compatibility!

frozenfrank (approved):
The code looks good to me. Good job making minimal changes that meet all the technical requirements previously discussed.

It sounds like there is still a discussion about the least confusing way to present this information on the UI to the students. We can discuss that, but it's small in comparison to bulk of this change.

DallinFromEarth:
Yeah I haven't had a chance to look at this since you @19mdavenport and I discussed the UI the other day. I just worry that putting too many numbers on the screen will confuse students about what the results really mean.

19mdavenport:
There are some more changes for the score views that could be improved in further PRs

PR #436, opened 2024-09-30
Provide contextualized note about potential issue with temporary code

frozenfrank:
## Overview
Heads up @DallinFromEarth! `UserController:loadCurrentReposIntoUpdateTable()` has an error-handling vulnerability.

I know this is temporary code, but it does have a vulnerability that may cause it to work improperly.

## PR Management
Feel free to close this PR if it's not a concern. Otherwise, build on this PR by pushing to the branch to resolve the concern.

Note that this PR is currently being merged into the `412-expand-url-passing-regex` branch, but this is only temporary. GitHub will automatically update it to merge into `main` when the other PR is merged.

### DIFF
--- a/src/main/java/edu/byu/cs/controller/UserController.java
+++ b/src/main/java/edu/byu/cs/controller/UserController.java
@@ -186,6 +186,10 @@ public class UserController {
                 Instant fakeUpdateInstant = mountainTime.toInstant();
 
                 if (updateDao.getUpdatesForUser(user.netId()).isEmpty()) {
+                    // NOTE: If one URL already updated to Canvas is found "invalid",
+                    // then the `cleanRepoUrl()` will send a **halt()** signal to Spark.
+                    // This could prevent later URLs from being submitted,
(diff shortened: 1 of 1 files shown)


19mdavenport:
This is only a problem if #417 is also merged in (which we probably would want)

frozenfrank:
@19mdavenport Yes, we want #417 to be merged in. Note that this PR is actually set to merge into that PR right now so those changes are, by design, going to be merged in with these notes.

frozenfrank:
This this does have a theoretical issue, we are operationally protected because of the following points:
1. The function in question only operates on data already successfully saved into our database
2. All the repo URLs in the database have been cleaned by our system already
3. The improved functionality accepts everything that was previously accepted (plus some more)
4. Therefore, there can be no data samples running through this system that would cause it to fail

However, to be safe, we may still address this issue during other refactors. If it doesn't get addressed soon, we can safely close this issue since we have proven the exclusive space of the potential error it pertains only to temporary code.

PR #439, opened 2024-10-17
Disable the stubs of integration tests

frozenfrank:
In the TA meeting today, we breezed over the 200+ lines of meaningful test and pointed out the two tests which have no content. Rookie mistake; my bad.

I didn't implement these tests because they actually function as integration tests and rely on several complex data structures being configured properly.

This PR marks the tests as `@Disabled` so that they do not provide false indicators of success.


ThanGerlek:
No worries man, we were just poking fun 😂

DallinFromEarth:
Thank you for your valiant service in testing the code James

DallinFromEarth (approved):

PR #452, opened 2024-10-28
Update README.md and split into several related files

ThanGerlek:
A PR for #451. Adds, updates, clarifies, and reorganizes "getting started" info for new devs joining the autograder team.

- Extract getting started info from `README.md`
- Reorganize information by use case (e.g. isolate Windows-specific info)
- Remove/replace outdated information
- Add relevant "getting started" info
- General improvements to setup instructions

PR #455, opened 2024-10-30
Frontend: ServerCommunicator

webecke:
Introducing... the **_ServerCommunicator_**!

The _ServerCommunicator_ reduces a large amount of duplicate code in setting up and handling `fetch` calls to access the Autograder backend.

## What it does
Almost all calls to the backend now use one of 3 methods available in _ServerCommunicator_: `getRequest`, `postRequest`, and `patchRequest`. A `doUnprocessedRequest` method is also available, but should only be used in very specialized cases, like the HonorChecker and Analytics tool.

The chief goal was to simplify calls to the backend. Previously, a call to the backend from a frontend service would look like this:
```
export const approveSubmissionPost = async (netId: string, phase: Phase, penalize: boolean) => {
    const response = await fetch(useAppConfigStore().backendUrl + '/api/admin/submissions/approve', {
        method: 'POST',
        credentials: 'include',
        headers: {
            'Content-Type': 'application/json'
        },
        body: JSON.stringify({
            netId,
            phase,
            penalize,
        })
    });

    if (!response.ok) {
        console.error(response);
        throw new Error(await response.text());
    }
}
```
Now that same call looks like this:
```
export const approveSubmissionPost = async (netId: string, phase: Phase, penalize: boolean) => {
    await ServerCommunicator.postRequest('/api/admin/submissions/approve',
      {
        netId,
        phase,
        penalize,
    })
}
```

## Details
Every request sends the user's authToken, and all requests ask for `application/json` content type.

All requests will log an error to the console if the Server returns a non-200 status-code, then throw a new `ServerError` with info about the error.

All request methods have an optional body parameter (except GET) which the _ServerCommunicator_ will automatically serialize into json for you. 

 _The rest of this section does not apply to `doUnprocessedRequest`_

The request methods are generics that will automatically parse the body of the server's response into the type specified as T in the call of the method.

By default, the _ServerCommunicator_ will throw an error if the Server does not return anything in the body of the response. This behavior can be stopped by passing `false` into the `expectResponse` parameter in the request call.

### ServerError
When the server returns a non-200 status code, and the ServerCommunicator throws an error, it is of type `ServerError`. It features member variables storing the endpoint called, the status code, and status text. 

Additionally, `ServerError` provides a handful of methods like `error.isUnauthorized` to allow other areas of the code to check for certain errors without being dependent on HTTP codes.

## Other Clean-up In This PR:
* Removed the unused `patch` endpoint for user updates.
* Removed the unused `ItemType` type.
* Removed the unused `SearchableList` component.

_Also we added @ThanGerlek to the about page. Welcome to the team!_


src/main/resources/frontend/src/App.vue
```diff
@@ -23,9 +24,13 @@ const logOut = async () => {
...
   } catch (e) {
-    alert(e)
+    if (e instanceof ServerError){
+      alert(e.message)
+    } else {
+      alert(e)
+    }
   }
-  router.push({name: "login"})
+  await router.push({ name: "login" })
```
frozenfrank, line 33: Good catch adding the missing `await` keyword!

src/main/resources/frontend/src/components/AboutPage.vue
```diff
@@ -91,6 +91,11 @@ onUnmounted(() => {
           <p class="tenure">Jan 2024-Dec 2025</p>
           <p class="contributions">Developed class-requirement enforcement systems, like late-days and git-commits</p>
         </div>
+        <div class="person">
```
frozenfrank, line 94: I'm kind of surprised we haven't already created a component to encapsulate the details of creating a new person. 

I'm imagining something like the following:
```vue
<Person
    name="Dallin Webeke"
    url="awesome.com"
    fa-icon="unicorn"
    tenure="Since forever ago"
    contributions="Robust Vue Management"
></Person>
```
or, a for loop that displays all the persons based on an in-memory (potentially dynamically loaded) list of contributors.

That doesn't need to be incorporated in time for this PR, but you can create a mental note (or an issue) if you agree.

src/main/resources/frontend/src/components/AboutPage.vue
```diff
@@ -91,6 +91,11 @@ onUnmounted(() => {
...
+          <p class="name">Than Gerlek <a href="https://github.com/ThanGerlek" target="_blank"><i class="fa-solid fa-not-equal"/></a></p>
+          <p class="tenure">Jan 2024-Apr 2026</p>
+          <p class="contributions">He's new to the team, but he's doing a lot</p>
```
frozenfrank, line 97: Glad to have @ThanGerlek on board!

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,70 @@
...
+  const response = await doUnprocessedRequest(method, endpoint, bodyObject)
+
+  if (!expectResponse) {
+    return null as T
+  }
+
+  // this makes sure there is actually something in the body before trying to parse
+  const text = await response.text()
+  if (text) {
+    return JSON.parse(text) as T
```
frozenfrank, line 37: As discussed in CS340, this move to interpret the data as `T` is convenient, but not safe. The backend could totally lie or break or change, or the front-end could be hitting the wrong endpoint, and the data could not actually line up with the type we expect it to be.

A less convenient but safer way to access this data would be by returning a type `unknown`. This would force users of the function to do an actual content check before typecasting it, which would not only give them an opportunity to fail more gracefully when it's wrong, but also be more confident that the data is actually correct.

It's not necessary to make this change, but I just want to make sure we're acknowledging the tradeoffs of this decision.

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+async function doRequest<T>(method: string,
+                            endpoint: string,
+                            bodyObject?: Object | null,
+                            expectResponse?: boolean): Promise<T> {
+  const response = await doUnprocessedRequest(method, endpoint, bodyObject)
+
+  if (!expectResponse) {
+    return null as T
```
frozenfrank, line 31: This is strange. Why are we typecasting `null` to `T` when it never will be?

We can express this relationship more clearly with TypeScript function overload signatures. Note that when signature overloads are provided in TS, the last function declaration is interpreted as the "implementation signature." The last signature must be compatible with all the previous "call signatures," and it is not exposed for public use.

```ts
function doRequest<T>(
    method: string,
    endpoint: string,
    bodyObject: Object | null,
    expectResponse: false,
): Promise<null>
function doRequest<T>(
    method: string,
    endpoint: string,
    bodyObject: Object | null,
    expectResponse: true,
): Promise<T>
function doRequest<T>(
    method: string,
    endpoint: string,
    bodyObject?: Object | null,
    expectResponse?: true,
): Promise<T | null>
function doRequest<T>(
    method: string,
    endpoint: string,
    bodyObject?: Object | null,
    expectResponse?: boolean,
): Promise<T | null> { /* implementation */ }
```

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,70 @@
...
+    body: bodyObject ? JSON.stringify(bodyObject) : null
+  });
+
+  if (!response.ok) {
+    console.error(`A ${response.status} error occurred while making a ${method} request to ${endpoint}`)
+    console.error(response)
+    throw new ServerError(endpoint, await response.text(), response.status, response.statusText)
+  }
+  return response
+}
```
frozenfrank, line 70: Watch those missing newlines at the end of files!

src/main/resources/frontend/src/network/ServerError.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  isBadRequest(): boolean { return this.status === 400; }
+  isUnauthorized(): boolean { return this.status === 401; }
+  isForbidden(): boolean { return this.status === 403; }
+  isNotFound(): boolean { return this.status === 404; }
+  isHonorCodeViolation(): boolean { return this.status === 418; }
+  isUnprocessableEntity(): boolean { return this.status === 422; }
+  isInternalServerError(): boolean { return this.status === 500; }
+}
```
frozenfrank, line 25: Trailing newline...

src/main/resources/frontend/src/services/adminService.ts
```diff
@@ -1,112 +1,44 @@
...
     try {
-        const response = await fetch(useAppConfigStore().backendUrl + '/api/admin/users', {
-            method: 'GET',
-            credentials: 'include'
-        });
-
-        return await response.json();
+        return await ServerCommunicator.getRequest<User[]>('/api/admin/users')
     } catch (e) {
         return [];
```
frozenfrank, line 9: I'm noticing that this `try ... catch ... return default value` is a common pattern among all these endpoints. A few of them go a step further by also logging an console message before returning the default value.

Can we create an overload or similar method that streamlines this process by also taking in a default value to respond to an error in the request? Potentially, the argument could look like: `onError: T | ((error: Error) => T)`. This would allow a user of the function to drop in a default value to return, **or** provide a function that would perform custom logic before returning a default value.

src/main/resources/frontend/src/services/authService.ts
```diff
@@ -30,16 +26,5 @@ export const loadUser = async () => {
...
-            'Content-Type': 'application/json',
-        },
-    });
-
-    if (!response.ok) {
-        console.error(response);
-        throw new Error(await response.text());
-    }
+    await ServerCommunicator.postRequest( "/auth/logout", undefined, false)
 }
```
frozenfrank, line 30: Watch those trailing newlines! @webecke 

https://github.com/softwareconstruction240/autograder/pull/452#issuecomment-2444578760

src/main/resources/frontend/src/services/configService.ts
```diff
@@ -37,7 +28,7 @@ export const setLivePhases = async (phases: Array<Phase>): Promise<void> => {
 }
 
 export const setCanvasCourseIds = async (): Promise<void> => {
-    await doSetConfigItem("GET", "/api/admin/config/courseIds", null);
+  await doSetConfigItem("GET", "/api/admin/config/courseIds", {});
```
frozenfrank, line 31: It looks like we have some whitespace inconsistency going on... Some of the lines in this file are intended with 2 spaces, but others are indented with 4. Just pick a value and stick with it.

Just we look into setting up a linter that will automatically correct these kinds of details whenever the code is saved?

src/main/resources/frontend/src/services/submissionService.ts
```diff
@@ -1,105 +1,41 @@
...
-    let url = useAppConfigStore().backendUrl + '/api/submission' + (phase === null ? "" : "/" + Phase[phase])
-    const response = await fetch(url, {
-        method: 'GET',
-        credentials: 'include'
-    });
-
-    if (!response.ok) {
-        console.error(response);
+    try{
+        return await ServerCommunicator.getRequest<Submission[]>('/api/submission' + (phase === null ? "" : "/" + Phase[phase]))
```
frozenfrank, line 7: See all of these methods doing the same `try ... catch ... return default value` pattern? As previously mentioned, a function would really do some good here.

src/main/resources/frontend/src/services/submissionService.ts
```diff
@@ -1,105 +1,41 @@
...
-    });
-
-    if (!response.ok) {
-        console.error(response);
-        throw new Error(await response.text());
-    } else {
-        return true;
-    }
+    await ServerCommunicator.postRequest("/api/admin/submissions/rerun")
 }
```
frozenfrank, line 41: Trailing newline

src/main/resources/frontend/src/services/userService.ts
```diff
@@ -1,43 +1,24 @@
...
-
-  if (!response.ok) {
-    console.error(response);
-    throw new Error(await response.text());
-  }
+const updateRepoPatch = async (repoUrl: string, endpoint: string): Promise<void> => {
+  await ServerCommunicator.patchRequest(endpoint, {
+    "repoUrl": repoUrl
+  }, false)
 }
```
frozenfrank, line 24: Trailing newline

frozenfrank (dismissed):
Overall, look at the glorious bonuses that come with removing code duplication! This is awesome!
image

I've left several comments about ways this PR could be improved even further. In my mind, I'd love to step in and _do_ some of the changes myself, but I haven't even finished my own PRs recently....

src/main/resources/frontend/src/network/ServerCommunicator.ts
frozenfrank: Alternatively, the main point of these changes could also be achieved with explicit TSDocumentation:
```ts

/**
 * Conveniently submits an HTTP Request and returns the intepreted result.
 * 
 * Only interprets and returns the response body when such a body is expected.
 * The response body is parsed into JSON, but it is not validated or guaranteed
 * to be the expected `T` type. The safest way to use this method is to set `<T=unknown>`.
 * 
 * @param method HTTP Method to submit
 * @param endpoint URL Endpoint to hit
 * @param bodyObject Optional. An object containing JSON body data to send with the Request
 * @param expectResponse Default: `false`. The response body will only be parsed when set to `true`.
 * @returns A {@linkcode Promise} resolving to the generic T (unsafely cast) when `expectResponse=true`, 
 *          and `null` all the rest of the time.
 */
function doRequest<T>(
    method: string,
    endpoint: string,
    bodyObject?: Object | null,
    expectResponse?: boolean,
): Promise<T | null>
```

frozenfrank: Additionally, the two approaches could be combined to result in the most explicitly documented code possible.

src/main/resources/frontend/src/services/submissionService.ts
webecke: A of the methods I replaced had the try, catch, console.error() the error. `ServerCommunicator` will now log the error automatically (but also throw something). I played around with making a `default value` parameter to the request methods, but I couldn't decide if that made the API too complicated. In some cases, it could look like
`ServerCommunicator.postRequest("/api/endpoint", { body }, false, { default empty})`. Would we want to make that default value a part of the request methods as an optional parameter? or a separate function?

src/main/resources/frontend/src/services/configService.ts
webecke: For the most part I've been matching whatever is already in the file without thinking about it. If we had a linter that would be great, especially since we don't actually have a style guide

src/main/resources/frontend/src/services/submissionService.ts
webecke: The python formatted that comes with IntelliJ pushes me to do trailing newlines at the end of files, and I've just picked it up there. When did we pick a style guide?

src/main/resources/frontend/src/services/adminService.ts
webecke: As I mentioned in another comment, I did play around with it. I'll go ahead and add back in the `defaultReturn` optional parameter

src/main/resources/frontend/src/network/ServerCommunicator.ts
webecke: Is this the crusade you're on now? 😂

src/main/resources/frontend/src/network/ServerCommunicator.ts
webecke: If this is a concern, I'm pretty sure we can loop through the json object `.keys()` and make sure it has all the members, before returning, and make it just a part of the request methods.

src/main/resources/frontend/src/network/ServerCommunicator.ts
webecke: I'll go through and document the new code. The whole `if (expectResponse) { return null as T }` shenanigan was the only way I could figure out how to get it to return null when no response was wanted. Ideally, if you didn't want a response, you would pass `null` or `void` in as the type. But that doesn't work in typescript since types don't really exist at runtime.

webecke: I'm a bit confused about what you are proposing with the overloads. How do the different signatures allow the user to get a null back?

src/main/resources/frontend/src/components/AboutPage.vue
webecke: Agreed!

webecke:
@frozenfrank thank you for your feedback and your help with overloads. I've added the overloads as we discussed, and I'll be adding documentation the next time I have time to work on this. _(Also, apparently as I'm typing this, vue type checks failed lol)_ So this PR still has more for me to do.

src/main/resources/frontend/src/network/ServerCommunicator.ts
frozenfrank: Bro don't "@" me! I'm just pointing out the glaring red circle that Git, GitHub, and the broader linux community have decided on ⛔️⛔️⛔️

src/main/resources/frontend/src/services/submissionService.ts
frozenfrank: See #459.

webecke:
@frozenfrank I've added docs now, with the help of my friend Claude. It seems that we have to have a lot of duplicated comments because of how the overrides work with JSDocs.

Also thanks for fixing the type error.

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,269 @@
...
+ * @template T - The type of the expected response (when expectResponse is true)
+ * @param {string} endpoint - The API endpoint to call
+ * @param {boolean} [expectResponse=true] - Whether to expect and parse a response
+ * @returns {Promise<T | null>} Promise that resolves to:
+ *   - The response data of type T when expectResponse is true
+ *   - null when expectResponse is false
+ * @throws {ServerError} When the request fails (meaning the server returned a code other than 2XX)
+ * @throws {Error} when expectResponse is true but no response is received
+ */
+function getRequest<T>(endpoint: string, expectResponse: false): Promise<null>;
```
frozenfrank, line 34: Note that the generic `<T>` is actually unused on this overload and can be removed.

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,269 @@
...
+ * @throws {Error} when expectResponse is true but no response is received
+ *
+ * @example
+ * // With response
+ * const user = await postRequest<User>('/api/users', { name: 'John' });
+ *
+ * // Without response
+ * await postRequest<void>('/api/logs', { event: 'action' }, false);
+ */
+function postRequest<T>(endpoint: string, bodyObject: Object | null, expectResponse: false): Promise<null>;
```
frozenfrank, line 76: Same here.

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,269 @@
...
+ * @throws {Error} when expectResponse is true but no response is received
+ *
+ * @example
+ * // With response
+ * const user = await patchRequest<User>('/api/users/123', { name: 'John' });
+ *
+ * // Without response
+ * await patchRequest<void>('/api/users/123/status', { status: 'active' }, false);
+ */
+function patchRequest<T>(endpoint: string, bodyObject: Object | null, expectResponse: false): Promise<null>;
```
frozenfrank, line 128: Same here.

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,269 @@
...
+ * @param {Object | null} [bodyObject=null] - The request body object to send (will be sent as JSON)
+ * @param {boolean} [expectResponse=true] - Whether to expect and parse a response
+ * @returns {Promise<T | null>} Promise that resolves to:
+ *   - The response data of type T when expectResponse is true
+ *   - null when expectResponse is false
+ * @throws {ServerError} When the request fails (meaning the server returned a code other than 2XX)
+ * @throws {Error} When expectResponse is true but no response is received
+ * @internal
+ */
+function doRequest<T>(
```
frozenfrank, line 174: And this one

src/main/resources/frontend/src/network/ServerCommunicator.ts
frozenfrank: I totally agree. I'm glad we were able to at least discuss and be aware of the consequences of this decision.

This is the same problem Dr. Wilkerson warned us about in our CS 340 class last week, but it will be acceptable to move in this direction.

src/main/resources/frontend/src/services/adminService.ts
frozenfrank: Any update on this?

frozenfrank (commented):
Great progress! Any update on the `defaultValue` parameter?

src/main/resources/frontend/src/network/ServerCommunicator.ts
webecke: Removed!

webecke:
> Great progress! Any update on the `defaultValue` parameter?

yep! per our offline discussion, I added a wrapper function called `getRequestGuaranteed` which just takes in an endpoint, and an object of type T called `errorResponse`. Should the server return a non 2xx code, or return nothing at all, the function will return the provided `errorResponse`. This has eliminated a lot of duplicated `try {} catch() {}` blocks

src/main/resources/frontend/src/network/ServerCommunicator.ts
```diff
@@ -0,0 +1,287 @@
...
+ * @param {string} endpoint - The API endpoint to call
+ * @param {T} errorResponse - The object the method call should return if the server
+ * returns nothing or responds with a non-2XX code
+ * @returns {Promise<T>} Promise that resolves to the response data of type T
+ */
+async function getRequestGuaranteed<T>(endpoint: string, errorResponse: T): Promise<T> {
+  try {
+    return await getRequest<T>(endpoint, true)
+  } catch (e) {
+    return errorResponse
```
frozenfrank, line 38: Great work!

Heads up: I notice that the documentation indicates an error will be logged, but I don't see the `catch` actually logging an error.

frozenfrank (commented):
Looking great!

src/main/resources/frontend/src/network/ServerCommunicator.ts
webecke: Yep! This is intentional. The `doRequest` method already logs the error before throwing a `ServerError`, so additional logging is not needed.

src/main/java/edu/byu/cs/server/Server.java
```diff
@@ -82,8 +82,6 @@ public static int setupEndpoints(int port) {
 
                 get("/users", usersGet);
 
-                patch("/user/:netId", userPatch);
-
```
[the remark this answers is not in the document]
webecke: removed!

19mdavenport (approved):

webecke:
@frozenfrank anything else you want changed?

PR #459, opened 2024-11-03
Add newlines to the end of all files...

frozenfrank:
## Application
This was applied automatically with the following regex:

Find: `([^\n])\z`
Replace: `$1\n`

The regex was applied to individual subdirectories so that the `node_modules/` directory 
and other target directories could be excluded from the transformation.

## Discussion
@webecke This is how I feel about trailing newlines :)

The questions is... How can I introduce automatic tooling that will handle this programmatically going forward?


webecke (approved):
Wow... you really do like these trailing new lines @frozenfrank 

LGTM

webecke:
> The questions is... How can I introduce automatic tooling that will handle this programmatically going forward?


I actually wanted to talk to you about that. I'm working on the `CONTRIBUTING.md` and what I've seen from examples online is a Style Guide. We don't currently have a style guide, so I think it would be good if we had some sort of tool that could at least help us out with new lines and tabs and such stuff.

PR #472, opened 2024-11-08
Simplify server communicator

frozenfrank:
Simplifies details pertaining to #455 which was merged before these could be included.

Read the commit descriptions for more information about each of the proposed changes.


webecke (changes requested):
It looks really cool, I like not having to use async everywhere. However, it breaks test student for some reason, triggering a 403 uncaught serverError when you try to access test student. I don't have time to look into it today, but as it stands this is broken.

frozenfrank:
@webecke Thanks for testing that. I'll try to spin it up myself and see what's going on (I haven't spun on the front-end in awhile though so hopefully this isn't difficult).

webecke (approved):
LGTM

PR #473, opened 2024-11-08
Migrate autograder to use Javalin

### LINKED ISSUE #443: Migrate to JavaLin to match upcoming change to Class Project
## Overview
Today in the TA meeting, it was announced that SparkJava is now unmaintained. The class project will be migrating to a fork of server called [JavaLin](https://javalin.io/).

We decided it would be meaningful to migrate the AutoGrader to using JavaLin as well.

This will probably require additional tests so we can migrate with confidence that things won't break.

## Approval
This was approved by @leesjensen and @jerodw

ThanGerlek:
Migrate the Server class to use the [Javalin](https://javalin.io/documentation) web framework instead of SparkJava. The logic transfers over fairly neatly but with a few changes noted below. Closes #443.

## Significant changes
- Spark's `Request` and `Response` objects are both replaced with a single `Context` object
- `Handler` methods (previously `Route`) can no longer have return values
- No `halt()` method exists
- `ctx.json(responseObject)` automatically serializes and sets the response body (using our Serializer class), and `ctx.bodyAsClass(ClassName.class)` deserializes the request
- Error responses are now handled via thrown Exceptions (ctx.status() works, but this is the standard to simplify/standardize/DRYify code). As such, various additional Exception classes have been added, largely corresponding to HTTP codes

## Dependencies

- #479
- #483 (which also depends on 479) is not a dependency, but it adds relevant unit tests, so it is recommended that it be merged in before this


ThanGerlek:
Still to do: service tests

src/main/java/edu/byu/cs/controller/WebSocketController.java
```diff
@@ -25,6 +26,7 @@ public static void onMessage(WsMessageContext ctx) {
         Session session = ctx.session;
         String message = ctx.message();
         String netId;
+        ctx.enableAutomaticPings(20, TimeUnit.SECONDS);
```
frozenfrank, line 29: Does this work properly even when the server is paused on a breakpoint, as if in a debug session?

My first guess is that the automatic pings would also be paused if the server is paused, which defeats the purpose of this timeout.

ThanGerlek: I did some testing, and you're right--that's something we'll have to be aware of when debugging. Fortunately it seems like the only time it affects anything is if the server hits a breakpoint partway through grading a submission, which probably won't happen often. Either way, the original intent was just if unit tests take a long time to run, and it does fix that.

frozenfrank: Is there a way to _also_ increase the timeout to a large value? That would be helpful for any server debugging, should that arise. Even if we, the autograder team, don't typically use this much, the students will need an increased timeout limit because they will be extensively debugging their servers.

ThanGerlek: I couldn't find one, and [this issue](https://github.com/javalin/javalin/issues/2329) from last month seems to suggest there isn't one. So I don't think so but I'm not sure 🤷

frozenfrank: > But students won't need it on the Autograder, will they? They're not debugging the code as its submitting to the autograder, the debug on their local machine, right?

@webecke My bad, let me clarify. I didn't mean to suggest that the students will be debugging on the AutoGrader. I meant to make reference to future semester where they will be building their chess projects using a version of Javalin. At that point they will be debugging code and stepping through websocket endpoints, and they will need to be able to keep their websocket connections alive.

> I couldn't find one, and https://github.com/javalin/javalin/issues/2329 from last month seems to suggest there isn't one.

@ThanGerlek Thanks for referring to that Javaline issue. [I just commented on it](https://github.com/javalin/javalin/issues/2329#issuecomment-2484739808) with my research on the WS timeout issue. From what I can tell _without having actually used Javalin myself_, the feature doesn't yet exist.

I see this as a symptom of using a premature product which is not yet fully featured. We should keep a list of this and any other weird things we discover about the platform. These kinds of details obviously don't make the entire decision, but the quantity and frequency and severity of multiple of them can influence our decision to look for a different platform as well.

We should take this is a reminder that this is just a community project. Upon reviewing the [recently merged PRs](https://github.com/javalin/javalin/pulls?q=is%3Apr+is%3Aclosed+is%3Amerged+-author%3Adependabot%5Bbot%5D), I see a total of 4 substantial PRs in the last 3 months. The professors suggested that this could be a good tool, but it may not be. We should keep our eyes and options open to switching to a different platform which may be better than Javalin.

src/main/java/edu/byu/cs/server/Server.java
```diff
@@ -23,99 +30,127 @@ public int start(int desiredPort) {
...
+                            // TODO does Javalin guarantee this...?
+                            // all routes after this point require authentication
+                            post("/logout", provider.logoutPost());
+                        });
 
-        before(provider.beforeAll());
+                        path("/api", () -> {
+                            before("/*", ctx -> {
+                                if (ctx.method() != HandlerType.OPTIONS) provider.verifyAuthenticatedMiddleware().handle(ctx);
+                            });
```
webecke, line 54: This verifyAuthenticatedMiddleware guarantees all the other endpoints are authenticated

frozenfrank:
@ThanGerlek Is this done? After we fix the merge conflicts, is there anything else that needs to happen to complete this PR?

ThanGerlek:
Just keep in mind that when merging, the auto-merge is probably insufficient--I usually had to manually confirm it did it correctly. Plus, any new endpoints will need to be added to Server, EndpointProvider, and the API unit tests.

src/main/java/edu/byu/cs/controller/UserController.java
```diff
@@ -4,68 +4,61 @@
...
         try {
             UserService.updateRepoUrl(studentNetId, repoUrl, adminNetId);
         } catch (BadRequestException e) {
-            halt(400, e.getMessage());
-        } catch (PriorRepoClaimBlockageException e) {
-            halt(418, e.getMessage());
-        } catch (InternalServerException e) {
-            halt(500, e.getMessage());
+            throw new BadRequestResponse(e.getMessage());
+        }  catch (InternalServerException e) {
```
frozenfrank, line 56: It looks like we got an extra white space here...

src/main/java/edu/byu/cs/server/SerializerAdapter.java
```diff
@@ -0,0 +1,25 @@
...
+import io.javalin.json.JsonMapper;
+import org.jetbrains.annotations.NotNull;
+
+import java.lang.reflect.Type;
+
+/**
+ * A Serializer class adapter which implements Javalin's JsonMapper interface.
+ * Allows Javalin's built-in serialization to use Serializer.
+ */
+public class SerializerAdapter implements JsonMapper {
```
frozenfrank, line 13: Amazing!!

src/main/java/edu/byu/cs/server/Server.java
frozenfrank, line 1: Is there a way to avoid the additional level of indentation introduced in this change? Shallower and flatter is better IMO

src/main/java/edu/byu/cs/server/Server.java
```diff
@@ -28,107 +33,136 @@ public int start(int desiredPort) {
...
+        })
 
-        init();
+        .exception(BadRequestException.class, haltWithCode(400))
+        .exception(UnauthorizedException.class, haltWithCode(401))
+        .exception(ResourceForbiddenException.class, haltWithCode(403))
+        .exception(ResourceNotFoundException.class, haltWithCode(404))
+        .exception(WordOfWisdomViolationException.class, haltWithCode(418))
+        .exception(UnprocessableEntityException.class, haltWithCode(422))
+        .exception(Exception.class, haltWithCode(500))
```
frozenfrank, line 151: 😍😍

src/main/java/edu/byu/cs/server/endpointprovider/EndpointProvider.java
frozenfrank, line 1: I see that the file existed before, but I can't tell what this purpose this file serves. Can help enlighten me?

src/main/java/edu/byu/cs/service/UserService.java
```diff
@@ -64,10 +68,10 @@ private static void setRepoUrl(String studentNetId, String repoUrl, String admin
         }
         if (historicalUpdate != null) {
             if (adminNetId != null) {
-                throw new PriorRepoClaimBlockageException("Repo is blocked because of a prior claim: " + historicalUpdate);
+                throw new WordOfWisdomViolationException("Repo is blocked because of a prior claim: " + historicalUpdate);
             } else {
                 LOGGER.info("Student {} was blocked from updating their url because of a prior claim: {}", studentNetId, historicalUpdate);
-                throw new PriorRepoClaimBlockageException("Please talk to a TA to submit this url");
+                throw new WordOfWisdomViolationException("Please talk to a TA to submit this url");
```
frozenfrank, line 74: 🐣🐰

src/main/java/edu/byu/cs/util/Serializer.java
```diff
@@ -31,6 +31,14 @@ public static String serialize(Object obj) {
         }
     }
 
+    public static String serialize(Object obj, Type type) {
+        try {
+            return GSON.toJson(obj, type);
+        } catch (Exception e) {
+            throw new SerializationException(e);
```
frozenfrank, line 38: Is this intended to be extend `RuntimeException`?

src/main/java/edu/byu/cs/server/endpointprovider/EndpointProvider.java
frozenfrank: Nvm. I now see that it enables testing by unifying a mock with the original interface.

frozenfrank (approved):
This looks like a great improvement to me.

I hope we can find a better way to help students with their debugging when they start using Javalin, but everything else looks much better. I appreciate seeing larger sections reduced of clutter because of better tools.

The comments I left are worth considering, but none of them are show stoppers to finish off this PR. Thanks @ThanGerlek for making this happen!

src/main/java/edu/byu/cs/util/Serializer.java
19mdavenport: I don't know. @ThanGerlek do you still care about this?

ThanGerlek: I think it already did, and I just never changed it. Y'all know my undying hatred of unchecked exceptions burns with the fiery passion of a thousand flaming meteorites, but I'm also not a dev anymore, so I don't really mind either way

Issue #480, opened 2024-11-14
Fullstack: refactor config storage and processing


webecke:
As we keep adding more and more config options, the management of config state on the front end is getting more and more difficult to manage. This needs to be refactored into something thats easier to manage and more expandable.

This could involve a restructuring into different config systems, or just a simplification of how config values are processed and stored. I'll figure it out once I get started on it.

webecke:
Potential idea: We make a ConfigDataTransferClass, and that just gets serialized when we ask for it on the front end. That way we can structure config things and nest them for sensible organization. Then, as long as we mirror it in the Config type on the front end, we can do minimal processing when we ask for the config. And it means we can have simpler additions to the config system.

ThanGerlek:
I'm not sure I understand the problem--is this a backend change, or a frontend refactor? The backend Config system certainly feels a bit messy (at least to me), but the frontend actually looks fairly well centralized. Especially given what we talked about in 340, about how a DTO and its corresponding Model class (the `Config` type and the config data store) should be separate--one is purely data, and the other has validation and logic. Am I missing what you're proposing?

webecke:
Currently, the way the front end storage is structured, to add a new config option, you have to set up three different things (all in the Config store, but its still a hassle, and as more and more things get added, it seems pretty messy).

This is really more of a full stack change, I want to make some changes to how the back end processes config and also how it sends it to the front end.

PR #482, opened 2024-11-16
Add configuration of penalty magic numbers

webecke:
This pull request introduces several changes to make penalty configurations more flexible and to improve logging and configuration management. The key changes include introducing new configuration parameters, updating the logic for calculating penalties, and adding new endpoints for updating penalty configurations. _Thanks copilot for the summary_

## Details
The following values are no longer hard coded, and are now configurable from the Admin Frontend
- Daily Late Penalty
- Max Late Days
- Clock Forgiveness
- Lines Required Per Commit
- Git Commit Penalty

Also in the PR are some functions to simply config parameter validation and processing.

## Testing
This PR doesn't add any new tests, but it runs fine with all the scorer and git verification tests. I did have to modify how the Memory Daos were loaded slightly to ensure default config values were initialized for the Memory Config Dao. Otherwise scorer tests would crash for not having penalty values set.

> [!CAUTION]
> When this changed is deployed to the server, all config options created in the PR will be initialized to zero. (Because of how the config system is set up) I would strongly recommend deactivating all phases BEFORE pushing the changes to the live system, so that an admin has time to go set up the correct values. The screenshot below illustrates the previously hard coded values
Screenshot 2024-11-16 at 13 28 37

## Future Work
As I've been working on the config system, I've realized several things are tedious or repetitive to work with. I created issue #480 to work on this.


19mdavenport (approved):

src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -33,11 +34,18 @@ public class Scorer {
...
     private final GradingContext gradingContext;
 
     public Scorer(GradingContext gradingContext) {
         this.gradingContext = gradingContext;
+        try {
+            ConfigurationDao dao = DaoService.getConfigurationDao();
+            PER_DAY_LATE_PENALTY = dao.getConfiguration(ConfigurationDao.Configuration.PER_DAY_LATE_PENALTY, Float.class);
+        } catch (DataAccessException e) {
+            LOGGER.error("Error while getting Per Day Late Penalty for Scorer.");
+            throw new RuntimeException(e);
```
frozenfrank, line 47: How far up the chain could this `RuntimeException` go? If the field were missing due to a misconfigurstion, would that result in crashing the autograder?

webecke: Also, because of how the config system was set up, in almost every cause it will return some sort of default value instead of throwing if the value wasn't set.

19mdavenport: That's a good point, James, I didn't see that. Dallin, you are incorrect of how far that exception would go. Might this be replaceable with a GradingException?

19mdavenport: Also, now that it's not a constant, might this be able to go closer to where it's actually used?

webecke: I'm sure we could do that. However, I might be wrong in the steps but I'm correct in its effect. I ran the code locally with a change that forced it to throw (just for testing) and nothing crashed, the grader just stopped.

webecke: Ok, so this still isn't breaking. We can rename the exception if it makes you happier, but the code is stable as is.

ThanGerlek: [Insert personal vendetta against unchecked exceptions here]

webecke: > [Insert personal vendetta against unchecked exceptions here]

I also am against unchecked exceptions. When I originally wrote the code, I had a bad assumption about when that line was run, so I thought a RunTimeException was appropriate. It wasn't until @frozenfrank asked about it that I specifically tested that and realized when it ran. Making it a checked exception would be great.

> In this case, it may be appropriate to introduce a new ConfigValueNotFound exception

I'm not sure this is the appropriate name or exception. If there's a DataAccessException thrown while accessing a config value, it truly is an issue with Data Access. All possible config keys exist in the database at startup (if they're new, they have a value of `default/changeme` and each class has some sort of default value. See #485) The only scenarios I can imagine this throwing is if the database gets disconnected, or someone drops the table while its running.

19mdavenport: @ThanGerlek was that meant as you though James and or myself had a vendetta or that you do?

ThanGerlek: No, only that I do. Sorry, lol

PR #483, opened 2024-11-19
Test server endpoints

ThanGerlek:
Adds a simple framework for testing Server endpoints and implements some basic tests.

## Significant changes

- TestServerFacade, a simple client
- MockEndpointProvider, for mocking Server endpoint routes (`Mockito.mock()` doesn't work on its own)
- Replaced HTTP PATCH calls with POST

## Dependencies

- #479


src/test/java/edu/byu/cs/server/ServerTest.java
```diff
@@ -0,0 +1,148 @@
...
+                Arguments.of("POST", "updateBannerMessage", "/api/admin/config/banner"),
+                Arguments.of("POST", "updateCourseIdsPost", "/api/admin/config/courseIds"),
+                Arguments.of("POST", "submitPost", "/api/submit"),
+                Arguments.of("POST", "adminRepoSubmitPost", "/api/admin/submit"),
+                Arguments.of("POST", "approveSubmissionPost", "/api/admin/submissions/approve"),
+                Arguments.of("POST", "submissionsReRunPost", "/api/admin/submissions/rerun")
+        );
+    }
+
+    // TODO figure out how to test PATCH calls... HttpURLConnection thinks it's an invalid method
```
webecke, line 63: Sooooooo I don't remember why I started using PATCH calls. I remember thinking I had a good reason, but I didn't write it down and the more we deal with it, the sillier it seems. I'd be ok going through the endpoints and switching them over to more conventional PUT requests

src/test/java/edu/byu/cs/server/ServerTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+    @AfterAll
+    static void stopServer() {
+        server.stop();
+    }
+
+    @BeforeAll
+    public static void init() {
+        server = new Server(mockedMockProvider);
+        int port = server.start(8080);
```
19mdavenport, line 78: Could this be port 0 instead of hardcoding 8080? (0 lets the OS choose any open port)

src/test/java/edu/byu/cs/server/exception/ResponseParseException.java
```diff
@@ -0,0 +1,7 @@
+package edu.byu.cs.server.exception;
+
+public class ResponseParseException extends Exception {
+    public ResponseParseException(String message, Exception cause) {
+        super(message, cause);
+    }
+}
```
19mdavenport, line 7: Do these need to be separate Exceptions? For the chess passoff jar, I did that so students could have a slightly better idea what's going on. Since this should only matter to TA's (and it's only be used by tests), would there be any problems with letting the TestServerFacade throw IOExceptions rather than custom test exceptions?

src/test/java/edu/byu/cs/server/ServerTest.java
ThanGerlek: Done. I overloaded Server.start() so it's only in one place as well.

src/test/java/edu/byu/cs/server/exception/ResponseParseException.java
ThanGerlek: I guess not? I'm not sure I would normally categorize server communication as being I/O, but I suppose it meets the definition, and it's probably better than one-time-use classes. I do want to catch and rethrow though for the sake of error message clarity

19mdavenport: Hmm. That's not exactly what I meant but this seems usable. If you want to stick with the catch and rethrow then one test-only exception might be warranted (if you want to, this looks fine too)

webecke (approved):
LGTM!

19mdavenport (approved):

frozenfrank:
What is preventing this PR from being merged? It looks like all issues have been resolved.

src/test/java/edu/byu/cs/server/ServerTest.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @AfterEach
+    public void tearDown() {
+        reset(mockedMockProvider);
+    }
+
+    @ParameterizedTest
+    @MethodSource("getEndpoints")
+    public void verifyEndpointCallsItsHandlersExactlyOnceInOrder(String method, String path, String endpointName) throws IOException {
```
frozenfrank, line 91: I like this clear and helpful code reuse to test all the endpoints!

frozenfrank (commented):
Looks good to me.

If you could summarize: what is being tested by these tests? It appears that the tests verify that the function _is called_, and returns a successful status code, but doesn't necessarily require specific data to be returned. Did I understand that correctly?

webecke:
So this has sat dormant for a month, is there a reason we haven't merged it? @frozenfrank @19mdavenport

webecke:
Huzzah! Thank you @frozenfrank

frozenfrank:
Huzzah! Thank you @ThanGerlek !!

webecke:
Actually, this seems to have broken something. Server no longer launches properly (in Server.java start() it should pass in 8080, but it doesn't anymore). We might need to rollback and double check this PR. @19mdavenport and I are looking into it right now.

ThanGerlek:
Rolling back [this commit](https://github.com/softwareconstruction240/autograder/commit/46d9eee2785a4be4a7b949fee8ac89da6f47f890) should fix that, I think

frozenfrank:
@ThanGerlek The startup issue was already fixed in #526

ThanGerlek:
Nvm then, I was just too slow 😂

webecke:
WE LOVE YOU @ThanGerlek COME BACK

ThanGerlek:
😢

PR #484, opened 2024-11-19
Use Javalin's built-in Response exceptions

ThanGerlek:
Replaces our inhouse HTTP Exception classes with Javalin's [default response](https://javalin.io/documentation#default-responses) exceptions. Javalin automatically catches these and sends the right status code, with the default message being the status code string (e.g. "Bad Request", "Server Error").

Obviously this depends on #473.

## Should we do this?

Pros:
- Much neater, shorter code
- Offloads Exception handling
- Prioritizes nicely with other .exception() clauses we might add

Cons:
- We have less control
- Greater dependency on Javalin
- Doesn't handle `null` messages well (see below)

#### Null messages

Throwing one of Javalin's default response exceptions with a `null` message string will simply return a nasty "Null value!" error message to the client (the no-parameter constructor works fine though). The `message` parameter in the constructor is annotated `@NotNull`, so it'll usually warn you if you try something like `throw new BadRequestResponse(null)`. However, this can be accidentally subverted with code like `catch (Exception e) {throw new BadRequestResponse(e.getMessage());}` (which happens in several places in the code) if the caught exception has no error message.


webecke (approved):
LGTM!

I'm going to cast my vote in favor of using Javalin's Exceptions. I get the argument against it, but the code looks nice this way and it means we don't have to worry about translating exceptions leaving the server. I'm not worried about the null thing, since we should be giving messages anyways with our exceptions. Great work!

19mdavenport (commented):
An argument against doing this is now the Services now depend on Javalin, which would be not ideal if we ever need to be switching to a different protocol or HTTP library (like switching spark to javalin). How likely do you think those would be going forwards?

ThanGerlek:
Probably not too likely I'd think, but either way it wouldn't be too hard to switch back to custom ones if we need to

src/main/java/edu/byu/cs/service/UserService.java
```diff
@@ -47,37 +45,36 @@ public static Collection<RepoUpdate> adminGetRepoHistory(String repoUrl, String
...
-            throw new InternalServerException("There was an internal server error in checking the repo update history for this url", e);
+            throw new InternalServerErrorResponse("There was an internal server error in checking the repo update history for this url");
         }
         if (historicalUpdate != null) {
             if (adminNetId != null) {
-                throw new WordOfWisdomViolationException("Repo is blocked because of a prior claim: " + historicalUpdate);
+                throw new ImATeapotResponse("Repo is blocked because of a prior claim: " + historicalUpdate);
             } else {
                 LOGGER.info("Student {} was blocked from updating their url because of a prior claim: {}", studentNetId, historicalUpdate);
-                throw new WordOfWisdomViolationException("Please talk to a TA to submit this url");
```
19mdavenport, line 80: Kinda sad to see this hilariously but appropriately named Exception go, but probably for the best. Do we want to keep the 418 status though?

ThanGerlek:
Will not implement. Turns out this would introduce deeper Service-level dependencies on Javalin than we expected, so we decided it's not worth it.

frozenfrank:
That decision is consistent with the principles Prof. Wilkerson covered today in CS 340. It should be pretty easy to convert our exceptions into Javalin exceptions, and I like the pattern of only performing this transformation when it leaves our system, and not within system.

Issue #466, opened 2024-11-07
Fullstack: schedule end of semester and block graded phases


webecke:
This work relies on changes in #465 

We need a way for TAs to schedule the Autograder to shut off the Live Phases of graded phases at the end of the semester. Mostly so @19mdavenport doesn't have to stay up until midnight to shut it off.

PR #485, opened 2024-11-19
add Instant default for config

webecke:
Some silly goose _me_ forgot to set up the default value of an Instant in the Configuration SQL dao. So that meant it would crash when starting up. This fixes that bug currently on the main branch that came from #466 

Guess I need to do more thorough testing


frozenfrank (approved):
Looks good. Typescript has ways to protect against this kind of error, but is there a way to identify it at compile time with Java?

ThanGerlek:
Not easily, it seems. The simplest way seems to be to simply overload all the relevant functions, which would let us get rid of that generic entirely, but that would mean modifying the ConfigurationDao interface and every relevant implementation whenever we add a new type.

PR #487, opened 2024-11-20
Fix submission table index

19mdavenport:
Codifies the fix to the "out of sort memory" error.


webecke (approved):
Gosh, I think this PR is a bit too complicated for a single PR, but I'll go ahead and approve it anyways

PR #491, opened 2024-11-26
Fix late penalty notes

19mdavenport:
Fixes the late penalty notes to not add anything if the submission is on time.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/score/Scorer.java
+++ b/src/main/java/edu/byu/cs/autograder/score/Scorer.java
@@ -429,11 +429,7 @@ public class Scorer {
 
         Integer maxLateDays = DaoService.getConfigurationDao().getConfiguration(ConfigurationDao.Configuration.MAX_LATE_DAYS_TO_PENALIZE, Integer.class);
 
-        if (numDaysLate >= maxLateDays)
-            notes += " Late penalty maxed out at " + numDaysLate + " days late: -";
-        else if (numDaysLate > 0)
-            notes += " " + numDaysLate + " days late: -";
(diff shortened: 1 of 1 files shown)


src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -429,11 +429,7 @@ public Submission generateSubmissionObject(Rubric rubric, CommitVerificationResu
 
         Integer maxLateDays = DaoService.getConfigurationDao().getConfiguration(ConfigurationDao.Configuration.MAX_LATE_DAYS_TO_PENALIZE, Integer.class);
 
-        if (numDaysLate >= maxLateDays)
-            notes += " Late penalty maxed out at " + numDaysLate + " days late: -";
-        else if (numDaysLate > 0)
-            notes += " " + numDaysLate + " days late: -";
-        notes += (int)(numDaysLate * PER_DAY_LATE_PENALTY * 100) + "% ";
```
webecke, line 436: what silly goose wrote this code?

src/main/java/edu/byu/cs/autograder/score/Scorer.java
```diff
@@ -465,6 +461,25 @@ else if (numDaysLate > 0)
         );
     }
 
+    private String makeLatePenaltyNotes(int numDaysLate, int maxLateDays, String origNotes) {
```
webecke, line 464: I like that its a separate function now, makes it more clear

webecke (approved):
LGTM

thanks for fixing my silly mistake

PR #497, opened 2024-12-17
Extracted Git Logic from UserService into GitHelper

### LINKED ISSUE #463: Extract git logic from UserService
## Overview

The `UserController` directly implemented a `git clone` command as a temporary filler while the real functionality was being implemented by another developer. The filler behavior should be replaced with a function call to the actual behavior, and optionally, the cloning behavior merged it and preserved as a requirement.

## Discussion
[UserService's isValidRepoUrl()] method seems like it doesn't belong in [that] class. Dealing with the specifics of `git` and cloning objects feels like it belongs in a different layer.

I would recommend moving this function into the `git/` directory (similar to having a separate DAO layer from the service layer).

_Originally posted by @frozenfrank in https://github.com/softwareconstruction240/autograder/pull/450#discussion_r1824728783_

## Specifics

This is the code in question:

https://github.com/softwareconstruction240/autograder/blob/967c395e24f4d12fb68a5bb6ea4821831cfcf91d/src/main/java/edu/byu/cs/service/UserService.java#L95-L107

Notice that, at a high level, this function is re-implementing the code already written in the `GitHelper.java` file. In fact, it was probably copy & pasted from there, and then tweaked for it's new environment.

https://github.com/softwareconstruction240/autograder/blob/967c395e24f4d12fb68a5bb6ea4821831cfcf91d/src/main/java/edu/byu/cs/autograder/git/GitHelper.java#L88-L100

_The simplest, although incorrect, resolution to this PR would be to replace the bulk of the body of `UserService:: isValidRepoUrl()` with a function call to `GitHelper::fetchRepo()`, and adapt the response._ That solution could look like this:

```java
    private static boolean isValidRepoUrl(String url) {
        File cloningDir = new File("./tmp" + UUID.randomUUID());
        try {
            // NOTE: Would require publicly exposing the method, and
            // allowing it to be called with an arbitrary URL (it currently
            // relies on the `GradingContext` object for the URL).
            GitHelper.fetchRepo(cloningDir, repoUrl);
            return true;
        } catch (GradingException ex) {
            return false;
        } finally {
            FileUtils.removeDirectory(cloningDir);
        }
    }
```

However, as discussed in https://github.com/softwareconstruction240/autograder/issues/463#issuecomment-2548107131, this direct replacement of functionality is not quite sufficient. In this case, we need to replace it with new code that performs it's original **intent**. 

The end result should really look like the following. Note that at this point, the `isValidRepoUrl()` function serves only as an adapter over the external `RepoUrlValidator.isValid()` method. It is a design choice to leave it as-is, or directly substitute `RepoUrlValidator.isValid()` into all (there is only one) occurrence within the `UserController.java` file. While there are reasons to leave it separate, in this case, it may make plenty of sense to go ahead and perform that substitution. (Note that this does not apply to the changing of the `RepoUrlValidator` API. That is not up for consideration here.)

```java
    private static boolean isValidRepoUrl(String url) {
        return RepoUrlValidator.isValid(url);
    }
```

## Scope of Changes

If this PR _only_ made the above change, it would be just fine and acceptable, and it would address the original issue. 

If the minimal required change is pursued, the following are expected to all be true:
 1. The direct `git clone` implementation is defined in exactly one location (`GitHelper`)
 2. Minimal changes are made to surrounding areas of the code
 3. No changes are made to public call signatures

However, we may be able to do better. If the developer so chooses, this issue could properly merge the two checks together so that a successfully validating a URL via the `RepoUrlValidator.isValid()` and `RepoUrlValidator.clean()` method performed the `git clone` command in addition to the current string parsing and in preparation for other future checks.

Since performing a `git clone` command dramatically changes the runtime characteristics of the function used in multiple places, a careful review would first need to be conducted to ensure that this doesn't break any existing assumptions or introduce a performance bottleneck. Specifically, the original string parsing was lightweight and fast, but a `git clone` performs multiple network calls and downloads the entire repository onto the machine. Some repos could be very large, or the internet could be down, and it could be that not all of the code was prepared for this. It is fine to convert this into a separate issue for development independent of this original issue.

If the merging behavior is pursued, the statements are expected to all be true, in addition to those shared earlier about the minimal required change:

4. The `isValid()`, and `clean()` methods of `RepoUrlValidator` are consistent with each other. By some method such as delegation, adapting, or calling the same functions, the pair of functions produce a corresponding successful or a failure message for all possible inputs.
5. The implementation of `RepoUrlValidator` follow the style of the original implementation. That is— the functionality is written as methods on an object, and a convenience `static` function adapts and exposes the for use external to the class.

## Intention

This is intended to give a clear, complete, and detailed overview of the state of this issue as well as the expectations for a PR that hopes to resolve this issue. If there remain ambiguities or questions about this issue, please post a clarifying question as a comment.

## Credit

Co-authored by @frozenfrank.

TheDavSmasher:
## Overview
Removed code duplicated from GitHelper and UserService where they would fetch a repo for different reasons but in the same manner. The code to do the repo fetching was made its own method.

RepoUrlValidator's isValid calls this method and does the minimal logic accordingly.

## Future Work
Work consolidating exception handling throughout the process, more importantly within UserService might be possible.

Closes #463

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -29,6 +29,7 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.UUID;
 
 public class GitHelper {
     private static final Logger LOGGER = LoggerFactory.getLogger(GitHelper.class);
@@ -89,6 +90,12 @@ public class GitHelper {
         gradingContext.observer().update("Fetching repo...");
 
         fetchRepoFromUrl(gradingContext.repoUrl(), intoDirectory);
+    }
+
+    public static File fetchRepoFromUrl(String repoUrl) throws GradingException {
+        File cloningDir = new File("./tmp" + UUID.randomUUID());
+        fetchRepoFromUrl(repoUrl, cloningDir);
+        return cloningDir;
     }
 
     //Method must be static due to the lack of GradingContext in cases where this method is required
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -49,7 +49,7 @@ public class GitHelper {
 
     public void setUp() throws GradingException {
         File stageRepo = gradingContext.stageRepo();
-        fetchRepo(gradingContext.stageRepo());
+        fetchRepo(stageRepo);
         headHash = getHeadHash(stageRepo);
     }
 
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -91,6 +91,7 @@ public class GitHelper {
         fetchRepoFromUrl(gradingContext.repoUrl(), intoDirectory);
     }
 
+    //Method must be static due to the lack of GradingContext in cases where this method is required
     public static void fetchRepoFromUrl(String repoUrl, File intoDirectory) throws GradingException {
         CloneCommand cloneCommand = Git.cloneRepository()
                 .setURI(repoUrl)
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -10,6 +10,7 @@ import edu.byu.cs.dataAccess.DaoService;
 import edu.byu.cs.dataAccess.DataAccessException;
 import edu.byu.cs.dataAccess.SubmissionDao;
 import edu.byu.cs.model.Submission;
+import edu.byu.cs.util.FileUtils;
 import edu.byu.cs.util.PhaseUtils;
 import org.eclipse.jgit.annotations.NonNull;
 import org.eclipse.jgit.api.CloneCommand;
@@ -31,6 +32,9 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.UUID;
 
+/**
+ * In this class, <pre>static</pre> methods are used to distinguish those that run are threadsafe and preserve no state.
+ */
 public class GitHelper {
     private static final Logger LOGGER = LoggerFactory.getLogger(GitHelper.class);
     private final GradingContext gradingContext;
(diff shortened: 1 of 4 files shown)


src/main/java/edu/byu/cs/autograder/git/GitHelper.java
```diff
@@ -88,8 +88,13 @@ private boolean shouldVerifyCommits() {
     private void fetchRepo(File intoDirectory) throws GradingException {
         gradingContext.observer().update("Fetching repo...");
 
+        fetchRepoFromUrl(gradingContext.repoUrl(), intoDirectory);
+    }
+
+    //Method must be static due to the lack of GradingContext in cases where this method is required
+    public static void fetchRepoFromUrl(String repoUrl, File intoDirectory) throws GradingException {
```
frozenfrank, line 95: It could be helpful to have an overload of this function that automatically generates the temp directory and returns the directory reference as the result.

It's optional, but it would be helpful now and could be helpful in the future too.

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
```diff
@@ -1,20 +1,27 @@
...
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class RepoUrlValidator {
 
     public static boolean isValid(@Nullable String repoUrl) {
+        File cloningDir = new File("./tmp" + UUID.randomUUID());
         try {
-            clean(repoUrl);
+            GitHelper.fetchRepoFromUrl(repoUrl, cloningDir);
```
frozenfrank, line 17: Close. We actually want to perform _both_ checks here to ensure the URL matches all requirements.

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
```diff
@@ -1,20 +1,27 @@
...
             return true;
-        } catch (InvalidRepoUrlException e) {
+        } catch (GradingException e) {
```
frozenfrank, line 19: I would recommend moving this closing logic into a separate function with the call signature: `static boolean canClone(String repoUrl)`. [Notice it doesn't throw an error.] It would be good to do a similar transformation to create a `static boolean canClean()` method.

Then the body of the `isValid()` method can be streamlined and improved in readability to `return canClean() && canClone()`.

frozenfrank (commented):
This is looking good, and is on the right track. I left a few comments, but this is nearly there. Good job!

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
TheDavSmasher: So do I return the clean check as well?
And, should the fetch use the return of clean or should clean be used by itself just as a check for exceptions?

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
TheDavSmasher: Can do

src/main/java/edu/byu/cs/util/RepoUrlValidator.java
TheDavSmasher: So then it'd be instead of catching exceptions, these two methods catch them and return Boolean, and then I return false if either of them is false?

TheDavSmasher: Also, written here or in githelper?

frozenfrank: Yes, you got it.

The two functions `canClean()` and `canClone()` effectively call other functions to do the work, but then interpret the result to always return a `boolean` value. In this case, they will be similar since they are both catching a specific kind of error to interpret as `false`; however, they are different because they are catching different errors. 

Additional checks in the future can also be introduced to return a `boolean` value, but they may not be catching errors (they may just be evaluating conditions). With the `isFork()` function that will be coming soon, it will make a network call, and then just report the result. Compare https://github.com/softwareconstruction240/softwareconstruction/issues/138.

This family of validation functions (`canClean()`, `canClone()`, and potentially `isFork()`)  belongs best in the `RepoUrlValidator` class. Here are a few reasons:
* They are each adapting logic from other places into a format that works well for the the `RepoUrlValidator.isValid()` function. Therefore, they belong with that function
* Notice that in most cases, the actual "work" does live in other classes and is being delegated to. That's a good thing. For example, the `canClean()` calls an instance method of the class, `canClone()` calls a method on a different class, and `isFork()` will ask a remote server for information and extract just the flag it cares about.
* Modifying the source locations for each of these isn't appropriate, or possible in all cases.

frozenfrank (approved):
Great work, @TheDavSmasher! Go ahead and press the "merge" button.

PR #498, opened 2024-12-23
ConfigurationSqlDao getValue type switch

TheDavSmasher:
## Overview
Method getValue previously ran an if-else chain over all the Types that are currently supported and then returned the string value parsed cast to the type. This was changed to a switch statement. 

## Details
Due to Java limitations, a switch statement cannot be run directly over a Class<?> object. Another option that failed was using reflection to create an instance and use Pattern Matching, but the types supported had inconsistent Constructor signatures (where a Constructor with the same arguments was not shared between all types).

Such being the case, I turned to a switch statement over the Class's name, which should provide the same readability improvement while keeping the same functionality.

## Testing
Tested manually with the Debugger's "Evaluate Expression" tool, using the same expressions/values as other usages within the application, as well as initializing the server (which utilizes this method).

## Additional Notes
This PR was made due to a code quality observation I made and is not linked to any issues for the moment (since I'm waiting on clarification/approvals on other issues).


frozenfrank (commented):
I like this change! I won't approve it just because I haven't specifically evaluated the code, but this code certainly looks like something we want to approve.

Nice work!

19mdavenport (approved):
Looks good to me

frozenfrank:
@TheDavSmasher When are you going to merge this?

TheDavSmasher:
Sorry I forgot, I'll do it on Monday

Issue #444, opened 2024-10-24
Docs: Improve Overall Documentation of AutoGrader Before Spring 2025


frozenfrank:
## Motivation

By Spring of 2025, all of the current contributors to the AutoGrader project will have graduated and left.

## Goals

We want to use the principles learned from other classes to leave behind meaningful documentation so that:
1. New TAs can get started developing and maintaining the AutoGrader
2. Existing TAs can understand how the system works at a high level
3. The project can continue and evolve to meet future needs of the CS 240 class and maybe be used in other contexts as well.

## Work List

| Abbv. | Meaning |
|--------|--------|
| SD | [Sequence Diagram](https://mermaid.js.org/syntax/sequenceDiagram.html) |
| PD | [Block Diagram](https://mermaid.js.org/syntax/block.html) _(Mermaid's best shot at a package diagram)_ |

This major project can be accomplished in steps which are represented as separate issues:
- #451
- #461
- #445 
    * Package Diagram
    * Sequence Diagram
- [x] Commit Verification
    * Commit Verification SD
    * Commit Verification PD
- [ ] External Gradebook (Canvas) Export SD
- [ ] Phase Graders SD
- [ ] Stopping & Starting System SD
- softwareconstruction240/softwareconstruction#150

## Resolutions

- #448
- #499
- #547
- #551
- #563
- #569
- #578
- #579

## Evolving Project
The deadline for this project is fixed, but the exact scope of it can change. 

Edit this Issue description to modify and adjust the items that should be included.

## Key Contributors
@19mdavenport @webecke @frozenfrank

webecke:
@19mdavenport is this issue done? If so, are we going to mark it closed?

I guess that might depend on our meeting later today.

frozenfrank:
Update: I attempted to organize this work using a [GitHub Milestone](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/about-milestones), but I found the "Milestones" feature to be underdeveloped for our purposes. Specifically, the feature doesn't support linking issues that span multiple repositories. Leverage this "Epic" style umbrella issue will be just fine for our purposes.

frozenfrank:
Thanks to everyone who helped contribute to the autograder and it's documentation. At this point, many PRs have addressed this. We'll count the rest of the ideas as "stretch goals" which could be revisited again in the future, but probably under different circumstances.

PR #440, opened 2024-10-17
422 backend commits with same timestamp should be downgraded in severity to warning

frozenfrank:
Resolves #422.

This PR:
* Will introduce tests to verify that the new functionality counts commits correctly
* Adjusts the language and severity of some common student `git` mistakes
* Preserves the mathematical correctness of the grading process and rigidity of requirements while also being forgiving of unintentional effects
* May refactor parts of the commit verification system to account for new kinds of data flows

PR #499, opened 2024-12-29
Document git commit verification

frozenfrank:
## Overview

Introduces a [mermaid sequence diagram](https://mermaid.js.org/syntax/sequenceDiagram.html) that visualizes and contextualizes the inner workings of the git commit verification system.

- This completes part of #444.

## Details

- Creates a sequence diagram with mermaid tools
- Introduces the diagram with notes about emphasis and intention
- Provides details about the key participants of the operation

## Testing
- [x] No testing needed. Documentation only.

## Dependencies

- This documents the latest architecture which was updated in #440.

## Future Work
- Provide high level overviews of the contents of the key `record` classes: 
    - `CommitVerificationResult`
    - `CommitsByDay`, especially the keys of `erroringCommits` map. Used in `CommitVerificationContext`.
- Provide tips for creating/extending/wrapping the `DefaultGitVerificationStrategy`
- Change the accumulation behavior of excluded commits to reside in strategy as opposed to being hard-coded in the base algorithm. This would make the algorithm more extensible and able to execute more arbitrary strategies.

## Additional Notes

- The sequence diagrams only appear when viewed within GitHub.
- We may consider moving this kind of documentation to a Wiki Page instead, but that isn't currently enabled
- This completes part of #444.
- We don't have a clearly defined naming convention for the `docs/` folder yet. I'm open to suggestions on the naming of the file.


webecke:
I like the idea of a wiki page, but it might make life more complicated.

The diagram looks great. I've never understood git commit verification, but it makes a lot more sense now. Great work!

frozenfrank:
> I like the idea of a wiki page, but it might make life more complicated.
> 
> The diagram looks great. I've never understood git commit verification, but it makes a lot more sense now. Great work!

@webecke Thanks for the helpful feedback! We'll just move forward without the WIKI page for now. 

Is there anything else you would like to see, or other improvements you would want before approving this PR?

19mdavenport (approved):
Nice Diagramming

PR #500, opened 2024-12-30
Admin Repo URL Submission HTTP Prefix

ghprime:
## Overview
 - This is my first PR to this repo, so it's small
 - If you submit a URL of a repo to be graded from the admin view, but don't prefix it with `http://` or `https://` it would allow you to submit but would fail. This felt strange to me since it does do some checking to see if the URL is potentially valid (namely making sure the URL contains `github.com`) before allowing submission, so it should either not allow the URL to submit without `http(s)://` or add the prefix in its absence.

## Details
 - Simply assures the existence of `http://` or `https://` at the beginning of the submitted url and adds it in its absence
 - Also I think that yarn updated some packages, but this is my first project with yarn so I'm not sure if that was me or that was done automatically

## Testing
- Tested manually
  - Valid URL w/o `http://` succeeds
  - Valid URL w/o `https://` succeeds
  - Valid URL w/ `http://` succeeds
  - Valid URL w/ `https://` succeeds
  - Invalid URLs still fail


src/main/resources/frontend/yarn.lock
```diff
@@ -99,11 +209,81 @@
...
+"@rollup/rollup-win32-ia32-msvc@4.14.0":
+  version "4.14.0"
+  resolved "https://registry.yarnpkg.com/@rollup/rollup-win32-ia32-msvc/-/rollup-win32-ia32-msvc-4.14.0.tgz#d7701438daf964011fd7ca33e3f13f3ff5129e7b"
+  integrity sha512-e/PBHxPdJ00O9p5Ui43+vixSgVf4NlLsmV6QneGERJ3lnjIua/kim6PRFe3iDueT1rQcgSkYP8ZBBXa/h4iPvw==
+
+"@rollup/rollup-win32-x64-msvc@4.14.0":
+  version "4.14.0"
+  resolved "https://registry.yarnpkg.com/@rollup/rollup-win32-x64-msvc/-/rollup-win32-x64-msvc-4.14.0.tgz#0bb7ac3cd1c3292db1f39afdabfd03ccea3a3d34"
+  integrity sha512-aGg7iToJjdklmxlUlJh/PaPNa4PmqHfyRMLunbL3eaMO0gp656+q1zOKkpJ/CVe9CryJv6tAN1HDoR8cNGzkag==
+
```
webecke, line 286: What's the deal with all of these new packages?

src/main/resources/frontend/yarn.lock
```diff
@@ -12,11 +12,121 @@
...
+"@esbuild/win32-ia32@0.20.2":
+  version "0.20.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/win32-ia32/-/win32-ia32-0.20.2.tgz#241fe62c34d8e8461cd708277813e1d0ba55ce23"
+  integrity sha512-HfLOfn9YWmkSKRQqovpnITazdtquEW8/SoHW7pWpuEeguaZI4QnCRW6b+oZTztdBnZOS2hqJ6im/D5cPzBTTlQ==
+
+"@esbuild/win32-x64@0.20.2":
+  version "0.20.2"
+  resolved "https://registry.yarnpkg.com/@esbuild/win32-x64/-/win32-x64-0.20.2.tgz#9c907b21e30a52db959ba4f80bb01a0cc403d5cc"
+  integrity sha512-N49X4lJX27+l9jbLKSqZ6bKNjzQvHaT8IIFUy+YIqmXQdjYCToGWwOItDrfby14c78aDd5NHQl29xingXfCdLQ==
+
```
webecke, line 129: What are all these new packages for?

webecke (commented):
This is a nice quality of life fix that is simple and well done. Seems to function to me! What's the deal with all the new packages?

src/main/resources/frontend/yarn.lock
ghprime: I'm not entirely certain, this is my first project using yarn. I thought they might have been added when I did the yarn install, maybe as a dependency for something? I'm not sure but I can remove this from the pull request

webecke (approved):
Thanks for cleaning that up!

LGTM!!!

frozenfrank:
Good job with the manual testing. I notice that no automated tested cases were added, is this because there is no system already in place to write automated front-end tests?

frozenfrank:
This change reverts other behavior that intentionally worked before.

The URL doesn't _have_ to be an HTTP* url because we also accept SSH urls of the form: `git@github.com/USERNAME/REPO`. GitHub sometimes shows that as the default URL when copying the link from the "clone repo" button depending on a users settings.

This change prevents use of these SSH urls. Read the `RepoUrlValidatorTest` file for more examples of URLS which should pass, and others that should fail.

https://github.com/softwareconstruction240/autograder/blob/9ea6e4d3ed4b3b9f86b59c939e5cea2897ecdc30/src/test/java/edu/byu/cs/util/RepoUrlValidatorTest.java#L67-L74

The real problem here is that we have the validation logic in two places. We cannot systematically verify that they will always be the same. A common approach is to have the real, complex validation on the server side while the front end only performs lightweight checking. Attempting to making the front end exactly match the server is futile since the rules while change. If possible, we would remove it from the front-end entirely, but there is a good reason to perform some basic validation in real time for the user before attempting to send it to the server.

I recommend reverting this change and putting it back to the way it was.

frozenfrank:
If we need to be able to accept URLS of the form "GitHub.com/username/repo" without the protocol, the solution is to adjust the back end URL parser to be able to interpret it.

ghprime:
This PR didn't break any existing functionality. Previously, the admin view _didn't_ support ssh urls. When I was going through the code to see where to put the fix, I came across this:
https://github.com/softwareconstruction240/autograder/blob/9ea6e4d3ed4b3b9f86b59c939e5cea2897ecdc30/src/main/java/edu/byu/cs/autograder/Grader.java#L60-L62 
which prevents the URL cleaning for some reason, so it wouldn't convert the ssh URL into an http URL. I assumed there was a reason behind it, so I decided to put the http:// part in the front end since the back end ignores any validation. 

In addition, in the admin view when we display the URL of the github repo we've submitted, it explicitly splits using `"://"` from `"http(s)://"` https://github.com/softwareconstruction240/autograder/blob/9ea6e4d3ed4b3b9f86b59c939e5cea2897ecdc30/src/main/resources/frontend/src/utils/utils.ts#L85-L87
so I assumed that we would only ever be worrying about github repos cloned using http. So if we wanted to support ssh urls, we would have to change a few things, which I'm not against doing. However, this PR doesn't break any existing functionality.

frozenfrank:
> This PR didn't break any existing functionality. Previously, the admin view _didn't_ support ssh urls. When I was going through the code to see where to put the fix, I came across this:
> 
> https://github.com/softwareconstruction240/autograder/blob/9ea6e4d3ed4b3b9f86b59c939e5cea2897ecdc30/src/main/java/edu/byu/cs/autograder/Grader.java#L60-L62
> 
> 
> which prevents the URL cleaning for some reason, so it wouldn't convert the ssh URL into an http URL. I assumed there was a reason behind it, so I decided to put the http:// part in the front end since the back end ignores any validation.
> In addition, in the admin view when we display the URL of the github repo we've submitted, it explicitly splits using `"://"` from `"http(s)://"`
> 
> https://github.com/softwareconstruction240/autograder/blob/9ea6e4d3ed4b3b9f86b59c939e5cea2897ecdc30/src/main/resources/frontend/src/utils/utils.ts#L85-L87
> 
> 
> so I assumed that we would only ever be worrying about github repos cloned using http. So if we wanted to support ssh urls, we would have to change a few things, which I'm not against doing. However, this PR doesn't break any existing functionality.

And thus we see the consequences of divergent and independent systems. Thanks for pointing out that no existing functionality was broken. You did a good job.

It is interesting because the URL cleaning tests specifically verify that the resultant URLs always have `https://` and never have `www.`, and they convert SSH URLs in to HTTPs URLs:

https://github.com/softwareconstruction240/autograder/blob/9ea6e4d3ed4b3b9f86b59c939e5cea2897ecdc30/src/test/java/edu/byu/cs/util/RepoUrlValidatorTest.java#L31-L47

Maybe this would be a good project— to properly send the URLs through cleaning while still allowing us admins to bypass whatever rule we were originally trying to bypass. (I'm not sure what that was anymore. If you don't find anything, feel free to invent and document a new and better approach.) @ghprime Do you want to convert this into an issue and tackle it?

PR #501, opened 2024-12-31
Enforce Prettier Code Quality

ghprime:
## Overview
Most of the front end was kinda a hodge podge of different styles of code, which while it doesn't affect how the code runs it does affect how new code will be written. We already have a `.prettierrc.json` file with code style in it, but nothing was enforced and it was obvious the formatter hadn't been run in a while. I decided to bring all the code into line by running the formatter on everything in the frontend and then enforce it going forward by adding a github action to check the formatting of the frontend code.

## Details
 - Formatted all the code in the frontend folder, with the exception of the code in `src/assets/fontawesome`
 - Added a github action to check the formatting of _just_ the frontend code

## Testing
 - Added a github action to test the code quality automatically when branches are merged into `main`

## Future Work
 - I decided to do this before doing any other work on the frontend to prevent more unformatted code and to have more structure going forward. This will be useful especially if we ever decide to incorporate the help queue into the autograder. I felt this wasn't going too far and doesn't need a lot of discussion since we already had a formatter set up.
 - That being said, I think that the Java code would also benefit from having a code formatter (solarlint comes to mind). However, unlike the frontend code that already had a formatter, the Java code currently doesn't have anything, so I didn't want to make a unilateral decision to add one, but I think that would be something to think about going forward, so I'm putting that into this section here to perhaps start the discussion.


webecke:
This looks exciting! Thank you for doing this. I'll be able to do a full review after the new year 👍

webecke (dismissed):
I'm thrilled about this PR. I don't agree with everything that prettier is doing in terms of formatting, but I'm just happy its making the format standard, so I'm willing to put up with it, especially since it will be done automatically. 

This PR doesn't seem to break anything. Although I personally am not able to test the prettier GitHub action, though it appears like it should work.

LGTM!

ghprime:
> I don't agree with everything that prettier is doing in terms of formatting, but I'm just happy its making the format standard, so I'm willing to put up with it, especially since it will be done automatically.

I'm not a huge fan of the prettier config we have either (I like having the semi colons for example) I didn't feel like changing the config since it was there before I was. That being said, it's not set in stone and if we wanted to change the config, right now would be the best time to do it. 

> This PR doesn't seem to break anything. Although I personally am not able to test the prettier GitHub action, though it appears like it should work.

The nice thing is that the github action was already run for this pr, you can see it in the 3 checks (2 from before and the 1 new one I added)

ghprime:
Before I merge this, I want a discussion on what format and therefore what config we want for our frontend code. We could keep the current one, but I think it would at least be beneficial if we talked a little about what we want. What do you think @webecke @19mdavenport @frozenfrank?

webecke:
I changed the config so that we have semi colons and trailing commas. I'm not exactly sure how much @19mdavenport and @frozenfrank will care.

I'm apathetic towards semi colons. I vote for trailing commas cause it makes things slightly easier when adding things to objects. Just my option though.

webecke:
So the prettier check failed, is that something I did wrong? The error just says we need to run prettier.

ghprime:
I also enforced double instead of single quotes, which is just a personal preference and I feel fits better with a java project since in java a single quote means a character (a type which js doesn't have) and double quote means a string (which is all js has). 

> So the prettier check failed, is that something I did wrong? The error just says we need to run prettier, isn't that what the action is doing?

The github action for this PR doesn't fix prettier issues, it just checks if there are any prettier issues. It's similar to running the test files; it won't fix any issues, only point them out. It's not that hard to change so that it would fix the issues and then commit the fix automatically, but I figured that if we had the programmer have to fix it manually it would perhaps have them write code that uses the new config going forward rather than rely on the action to fix the code for them automatically. 

In addition, if the pr committed automatically, then it might lead to the programmer forgetting to pull their branch if they need to make changes to it and leading to merge conflicts. 

I'm not entirely opposed to having the github action automatically fix the prettier issues, but I feel that having the github action serve as a check rather than an autofixer would be better.

webecke (approved):
I think this is wonderful! Thank you for helping me understand whats going on. I say go for it, thanks for putting this together!

LGTM!

PR #502, opened 2025-01-04
Revert "Admin Repo URL Submission HTTP Prefix"

frozenfrank:
Reverts softwareconstruction240/autograder#500.

See https://github.com/softwareconstruction240/autograder/pull/500#issuecomment-2571401754 for a discussion on why this is necessary.


ghprime:
See https://github.com/softwareconstruction240/autograder/pull/500#issuecomment-2571410644 for why this is unnecessary.

19mdavenport:
Is this still something that you want to be merged? It seems like in the comments under #500 that may not be the case

frozenfrank:
To be clear, I do not think #500 took the right approach to resolving the issue identified.

Rather than requiring an HTTP URL, it would have been much better to make it work when the protocol isn't provided.

However, after understanding that existing functionality wasn't _broken_, I now understand that simply reverting the change has little effect. I'll withdraw this PR.

We should revisit this issue and properly address it. A full resolution has been contemplated already in some the comments of related items.

We can keep the band-aid fix until someone is willing to actually operate on the patient.

PR #518, opened 2025-01-17
Backend: TrafficController publicly exposes sessions hashmap

### LINKED ISSUE #475: Backend: TrafficController publicly exposes `sessions` hashmap
## Overview
The `TrafficController` class publicly exposes a data structure. It should be privately managed and exposed only through authorized methods.

This issue improves adherence to the data-hiding principle and also promotes loose coupling between sub-systems.

## Problem
```java
    /**
     * A map of netIds to sessions that are subscribed to updates for that netId
     */
    public static final ConcurrentHashMap<String, List<Session>> sessions = new ConcurrentHashMap<>();
```

## Solution
Mark the variable as `private`. Provide helper methods to surface the functionality needed by other classes. Update the usages.

Maillman:
## Overview
Resolves #475.
The sessions hashmap was publicly exposed. I simply mark the variable as private, provided a helper method called `getSessions()` to surface the functionality needed by other classes. and updated the usages. Let me know if there are any changes I need to make.


frozenfrank (changes requested):
Marking it `private` is a start, but returning the entire object as `ConcurrentHashMap<String, List<Session>>` doesn't achieve the goals of hiding data. Returning the reference to the object still allows other places to manipulate it potentially unexpected ways.

I would expect the needs for that object to be decomposed and expressed in multiple helper functions.

src/main/java/edu/byu/cs/controller/TrafficController.java
```diff
@@ -22,20 +22,21 @@ public class TrafficController {
...
     public static boolean hasSession(String netId, Session session){
-        List<Session> sessionList = sessions.getOrDefault(netId, Collections.emptyList());
-        return sessionList.contains(session);
+        List<Session> sessionList = sessions.get(netId);
+        return sessionList == null || sessionList.contains(session);
     }
 
     public static void addSession(String netId, Session session){
-        List<Session> sessionList = sessions.getOrDefault(netId, Collections.emptyList());
+        sessions.putIfAbsent(netId, new ArrayList<>());
```
frozenfrank, line 38: What do you think about replacing this line with `addNetId(netId)`? That would reuse the existing code.

src/main/java/edu/byu/cs/controller/TrafficController.java
```diff
@@ -22,20 +22,21 @@ public class TrafficController {
     private static final ConcurrentHashMap<String, List<Session>> sessions = new ConcurrentHashMap<>();
 
     public static void addNetId(String netId){
-        sessions.put(netId, Collections.emptyList());
+        sessions.put(netId, new ArrayList<>());
```
frozenfrank, line 25: What do you think about replacing `sessions.put()` with `sessions.putIfAbsent()`? It would mean that repeated calls to `addNetId()` would not wipe out existing sessions.

Without studying the usage of the function myself, I can't point out specific usage examples. However, I can imagine that it may have an implication if the same student were to login on from multiple browser tabs at the same time.

src/main/java/edu/byu/cs/controller/TrafficController.java
```diff
@@ -22,20 +22,21 @@ public class TrafficController {
...
 
     public static List<Session> getSessions(String netId){
         return Collections.unmodifiableList(sessions.getOrDefault(netId, Collections.emptyList()));
     }
 
     public static boolean hasSession(String netId, Session session){
-        List<Session> sessionList = sessions.getOrDefault(netId, Collections.emptyList());
-        return sessionList.contains(session);
+        List<Session> sessionList = sessions.get(netId);
+        return sessionList == null || sessionList.contains(session);
```
frozenfrank, line 34: This logic seems incorrect to me. I think it should be:

```java
return sessionList != null && sessionList.contains(session);
```

src/main/java/edu/byu/cs/controller/TrafficController.java
```diff
@@ -24,8 +21,30 @@ public class TrafficController {
...
+        return sessionList.contains(session);
+    }
+
+    public static void addSession(String netId, Session session){
+        List<Session> sessionList = sessions.getOrDefault(netId, Collections.emptyList());
+        sessionList.add(session);
+    }
+
+    public static void clearSessions(String netId){
+        sessions.remove(netId);
```
frozenfrank, line 44: Should we call `session.close()` on any sessions in the data structure before we drop it? That may prevent memory leaks and lead to more correct behavior in corner cases.

I know that wasn't happening before, but this could be a way we would be improving and stabilizing the system during the refactor.

src/main/java/edu/byu/cs/service/SubmissionService.java
```diff
@@ -56,7 +56,7 @@ private static void startGrader(String netId, Phase phase, String repoUrl, boole
         QueueItem qItem = new QueueItem(netId, phase, Instant.now(), false);
         DaoService.getQueueDao().add(qItem);
 
-        TrafficController.getSessions().put(netId, new ArrayList<>());
+        TrafficController.addNetId(netId);
```
frozenfrank, line 59: Is this call still necessary? I suspect that the existence of `netId` may currently carry special meaning, but it doesn't seem like a good practice to me. I would rather rely on the `putIfAbsent()` behavior already implemented in the `TrafficController`.

@webecke @19mdavenport Is there any reason the `SubmissionService` class _needs_ to be interacting directly with the `TrafficController`? This seems like something that may not be necessary.

src/main/java/edu/byu/cs/controller/WebSocketController.java
frozenfrank, line 1: These improvements look much better. Thank you!

frozenfrank (changes requested):
This is looking much better! Great work!

In the comments, I've proposed a few improvements and spotted a few items that may not be working as intended. Let's keep working through those.

This is shaping up to be a very helpful. Thanks for the contributions!

src/main/java/edu/byu/cs/service/SubmissionService.java
19mdavenport: @frozenfrank I believe it's necessary to put a netid in somewhere before grading starts. Do you have a suggestion of somewhere else to put it?

src/main/java/edu/byu/cs/controller/TrafficController.java
19mdavenport: Oops. That's my fault (I was with @Maillman in person)

src/main/java/edu/byu/cs/service/SubmissionService.java
frozenfrank: That's exactly what I'm wondering. _Why_ does the `netid` need to be in the TrafficController before grading starts? Why isn't that already handled the SubmissionService as it's registering the session initially?

19mdavenport: I believe having the netid be in the map is how the WebSocketController knows that someone is the the grading queue or being graded

Maillman: Commit [366649c](https://github.com/softwareconstruction240/autograder/pull/518/commits/366649cf03139375bc6a91ac4a1b048abe325d7a) now removes this call and the WebSocketController now checks for the `netid` in the grading queue using the queueDAO instead of the map in TrafficController.

19mdavenport:
This looks good to me whenever @frozenfrank is satisfied with it

src/main/java/edu/byu/cs/controller/WebSocketController.java
```diff
@@ -25,27 +26,67 @@ public void onError(Session session, Throwable t) {
...
+    /**
+     * Indicates whether the student is the pass-off queue.
+     *
+     * @param netId The student NetId to verify.
+     * @return A boolean indicating the result.
+     * @throws RuntimeException When a {@link DataAccessException} occurs.
+     */
+    private boolean isInQueue(String netId) {
+        try {
+            return DaoService.getQueueDao().isAlreadyInQueue(netId);
```
frozenfrank, line 77: After we switched to checking the QueueDao (which was definitely a great call), the method which was already kind of long and complex became too long and complex for me. I went ahead and extracted several operations into helper functions to improve the readability and documentation of the overall process.

Now the main algorithm represented in the `onMessage()` method is much more understandable!

frozenfrank (approved):
Thanks for making all these changes, @Maillman! These are great.

I especially love how you dug a little deeper to figure out a better way to handle the registration validation. Switching to the `QueueDao` in 366649cf was the right way to go.

After that change, it became prudent to refactor the surrounding method to break up the complex `if` and `try`/`catch` structures and improve the overall readability. I went ahead and did that.

At this point, there are no other blockers. Go ahead and _merge_ this whenever you like!

Issue #510, opened 2025-01-12
Fullstack: Allow holidays to be updated dynamically with the config system


webecke:
Currently, every year, the list of holidays must be updated in the code. Let's at least make this something dynamically done in the config system.

https://github.com/softwareconstruction240/autograder/blob/185718bfe36328eda255d60b6a6903f9a875cd42/src/main/java/edu/byu/cs/autograder/score/LateDayCalculator.java#L263-L267

frozenfrank:
Technically, we could fill in the holiday dates for many years at a time. The current academic calendar gives the dates of holidays all the way through 2028. However, it felt more prudent to make the thing configurable rather than letting the maintainers in 2028 figure out the surprise ;)

frozenfrank:
See #219 where the code was introduced. There are two acceptable formats:
1. Dense, single-line format
2. Readable, multi-line format with comments

Both can be stored as a string in the database, but they would likely be interacted with differently. The dense format would likely be the result of picking dates from a calendar. The multiline format could be edited directly by users and annotated to provide context.

To me, it seems easier to provide a long text field and allow pasting or editing the holidays dates directly. However, it would likely be a better experience if editors could choose multiple dates from a year-view calendar etc... to toggle them on/off as holidays.

In the end, both are acceptable. This is a relatively infrequent task to complete, but it is one of the required items of maintenance for the autograder. While there are two acceptable formats for the holiday system, it goes without saying that any other format can be converted into an acceptable ease with minimal effort if that makes the most sense for this issue.

PR #521, opened 2025-01-21
511 verify future holidays are configured

### LINKED ISSUE #511: Verify Future Holidays are Configured
https://github.com/softwareconstruction240/autograder/blob/185718bfe36328eda255d60b6a6903f9a875cd42/src/main/java/edu/byu/cs/autograder/score/LateDayCalculator.java#L149-L151

_Originally posted by @frozenfrank in https://github.com/softwareconstruction240/autograder/pull/219#discussion_r1912702595_

frozenfrank:
## Overview

Resolves #511. Inserts checks so that academic holidays can be relied on to be up to date.

## Details

This provides a code-level, hard stop detection assertion to verify that academic holiday configuration is not stale. This is a redundant layer to the one @webecke will complete in a more admin friendly manner in #510.

This change will cause errors to be thrown and refuse to perform any grading when the holidays are incorrectly configured. This should enable an admin to quickly identify the error during smoke testing at the beginning of a semester. Helpful error messages are thrown that explain the problem clearly to an admin user.

Hopefully, these checks never need to throw errors, but when they do, it will be a helpful and important reminder that will keep the system running smoothly. This is one adjustment that will hopefully make the AutoGrader easier to maintain in the future.

Test cases are included to protect against regressions of this behavior.

## Testing

- [x] Tested edge cases

### DIFF
--- a/src/test/java/edu/byu/cs/autograder/score/MockLateDayCalculator.java
+++ b/src/test/java/edu/byu/cs/autograder/score/MockLateDayCalculator.java
@@ -7,6 +7,10 @@ import java.time.ZonedDateTime;
 public class MockLateDayCalculator extends LateDayCalculator {
     private final LateDayContext fixedContext;
 
+    public MockLateDayCalculator() {
+        this(0);
+    }
+
     public MockLateDayCalculator(int daysBeforeDueDate) {
         super();
         var now = ZonedDateTime.now();
@@ -16,5 +20,11 @@ public class MockLateDayCalculator extends LateDayCalculator {
     @Override
     protected LateDayContext doFetchLateDayContext(Phase phase, String netId) {
         return fixedContext;
+    }
+
+    @Override
+    protected void assertFutureHolidaysConfigured() throws RuntimeException {
+        // Disable these checks.
+        // Do not require tests to be chronologically up-to-date.
     }
 }
--- a/src/test/java/edu/byu/cs/autograder/score/MockLateDayCalculator.java
+++ b/src/test/java/edu/byu/cs/autograder/score/MockLateDayCalculator.java
@@ -12,7 +12,7 @@ public class MockLateDayCalculator extends LateDayCalculator {
     }
 
     public MockLateDayCalculator(int daysBeforeDueDate) {
-        super();
+        initializePublicHolidays(); // Initialize as blank; without reading default values
         var now = ZonedDateTime.now();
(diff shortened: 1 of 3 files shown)


webecke (approved):
A very simple change that will provide some operational safety to future TAs.

LGTM

src/test/java/edu/byu/cs/autograder/score/MockLateDayCalculator.java
```diff
@@ -17,4 +21,10 @@ public MockLateDayCalculator(int daysBeforeDueDate) {
     protected LateDayContext doFetchLateDayContext(Phase phase, String netId) {
         return fixedContext;
     }
+
+    @Override
+    protected void assertFutureHolidaysConfigured() throws RuntimeException {
+        // Disable these checks.
+        // Do not require tests to be chronologically up-to-date.
+    }
```
Maillman, line 29: Is this function something that's planned to be implemented in the future?

Maillman (approved):
I am a little curious about the empty assertFutureHolidaysConfigured() method; but other than that, I think it looks fine.

src/test/java/edu/byu/cs/autograder/score/MockLateDayCalculator.java
frozenfrank: @Maillman Great question!

The method you highlight is in the `MockLateDayCalculator` class. Generally speaking, this class extends `LateDayCalculator`, but modifies it's behavior for the purposes of testing.

In this particular case, the method is `@Override`ing this method from the base class:

https://github.com/softwareconstruction240/autograder/blob/3ef0ce5f908defa1f0b91ddc1667c9bcc0cdc946/src/main/java/edu/byu/cs/autograder/score/LateDayCalculator.java#L299-L325

The method is the one introduced in this PR. It requires that the dates in the holiday config are current with the present time, and throws warnings when they are out of date.

The method is _only_ called from one place, but it is called any time holidays are configured in the class:

https://github.com/softwareconstruction240/autograder/blob/3ef0ce5f908defa1f0b91ddc1667c9bcc0cdc946/src/main/java/edu/byu/cs/autograder/score/LateDayCalculator.java#L294-L297

 If we were to do nothing, then the tests would stop working all of the sudden when the current date passed the dates they have hard-coded for the purposes of testing. This is not a helpful feature. We want the tests to always work, but still have the guarantee that the actual production system is configured properly.

Therefore, the `MockLateDayCalculator` overrides the method with a separate version that does nothing. These checks will **never evaluate** in the tests.

PR #524, opened 2025-01-23
Refactor Configuration Communication

webecke:
## Overview
The communication of configuration values between the front end and the backend felt sloppy. This is my (probably amateur) attempt to simplify it. 

## Details

- Public and Private Config Transfer Objects
  - The front end and the back end both have matching types/classes that are used to transfer configuration values from the backend to the frontend.
  - This assists in the serialization and standardization of new configuration values
-  Course IDs config
   - Previously, the process to change the course ID required you to enter an id, then submit, then click a different button to load in the rubric and assignment ids. This PR changes it so those ids are loaded simply by entering the course ID
   - Course IDs are rejected and admin user is notified if the Autograder doesn't have access to that course

## Testing

- [ ] Tested edge cases
I tested the functionality of each config value. However, we don't have a testing framework for the front end yet. That is something we should work on soon.

## Additional Notes

I would appreciate any eyes on this that would be willing to give this a look. I think the system of config communication I've built makes sense, but I could be missing something or done something poorly.


webecke:
This had a lot of merge conflicts, and auto-merge things that broke things. However, it is now ready to go and is up to date with main.

It looks scarier than it is. Many of the 29 files changed are just refactoring of accessing config variables on the front end. Instead of `useAppConfigStore().shutdownSchedule` its now `config.private.shutdown.schedule`. Config values are nested and grouped with related config items.

19mdavenport (approved):
LGTM

PR #526, opened 2025-01-23
fix endpoint testing bug that prevents server from starting correctly

webecke:
## Overview

#483 broke the way the server starts up, preventing the front end from finding the backend. This fixes it.

## Details

- Re-states that the backend needs to run on 8080
- Tells the test package to run on port 0, which seems to let it run

## Testing

- [X] Added/updated unit tests
- [ ] Tested edge cases

## Additional Notes

>[!Important]
> This is to fix a bug that prevents the proper functioning of the Autograder. It must be merged before the next deployment

### DIFF
--- a/src/main/java/edu/byu/cs/server/Server.java
+++ b/src/main/java/edu/byu/cs/server/Server.java
@@ -18,7 +18,7 @@ public class Server {
     }
 
     public int start() {
-        return start(0);
+        return start(8080);
     }
 
     public int start(int desiredPort) {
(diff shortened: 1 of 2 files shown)


19mdavenport (approved):

frozenfrank:
@webecke Thanks for catching this 🙊

Is there a way our automated tests could have caught this sooner (lol)?

webecke:
> Is there a way our automated tests could have caught this sooner (lol)?


As the testing infrastructure stands now, I'm not sure how it would be done. We probably need to make an issue to create full length integration test from front to backend.

frozenfrank:
@webecke The new issue has been created.

Thank you for the quick response and resolution to the problem.

Issue #527, opened 2025-01-24
Create Server Integration Test


frozenfrank:
> Is there a way our automated tests could have caught this sooner (lol)?

As the testing infrastructure stands now, I'm not sure how it would be done. We probably need to make an issue to create full length integration test from front to backend.

_Originally posted by @webecke in https://github.com/softwareconstruction240/autograder/issues/526#issuecomment-2611256106_

PR #529, opened 2025-01-27
Fix bug where repo change requests failed

webecke:
## Overview

It was discovered that on the current main branch, trying to change the repo of yourself as a student would result in a network error. This fixes that.

## Details

- Direct front end to call repo update endpoints as a `POST` request
- Remove unused `PATCH` request methods from `ServerCommunicator`
- Rename `UserService` methods to not reference the HTTP method used

## Testing

- [ ] Tested edge cases

## Suspected Cause
When the great and late and beloved @ThanGerlek made his changes with #483 with the endpoint provider and whatnot, he changed the `/api/repo` endpoint to be a `POST` request. However, he didn't tell the front end about this change, and thus we got a 405 error.

## Future Work

When @frozenfrank and others previously talked about making frontend tests and regression tests for the API, I understood they were needed, but couldn't be bothered. After the second such bug in a week, I am now fully convinced about the necessity of these frontend/integration tests and have moved #527 to the top of my to-do list.

### DIFF
--- a/src/main/resources/frontend/src/components/RepoEditor.vue
+++ b/src/main/resources/frontend/src/components/RepoEditor.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { isPlausibleRepoUrl } from "@/utils/utils";
 import { defineEmits, reactive } from "vue";
-import { adminUpdateRepoPatch, studentUpdateRepoPatch } from "@/services/userService";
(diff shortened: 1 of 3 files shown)


ZakkeryDaRebel (approved):
I ran the code and had it work, I was able to make my repository be accepted, so I approve of this.

Issue #442, opened 2024-10-22
CommitVerification: Reduce Undue Penalties on Students Submitting Early


frozenfrank:
## Overview
Students passing off early face undue, unnecessary-positive Commit Verification warnings if they haven't committed on enough days.

**Give students credit towards their "required days with commits" for each day early that they pass off.**

## Discussion
The main point of the "day" requirement of the commit verification system is to ensure that students are not procrastinating work and doing it all at once.

We already have an informal policy that students passing off early are not subject to this requirement since they have clearly already worked ahead. The thinking goes that, if they have already completed the project now— before it's due—, then they could easily add an empty commit tomorrow and another empty commit the day after and submit with no issues.

However, rather than rewarding students for working ahead, we end up inadvertently penalizing these students if they complete the phase early, but in one day.

## Example
> **Student**: Hey TA, I got this when I tried to submit my code just now. I made sure that my code was spread out around 12 to 15 pushes but I forgot that the pushes had to be spread out over 3 days too. I wrote all the code first and then started pushing piece by piece.
> **Student**: It says I need to talk to a TA to receive the grade. Could you please help me out with this :sob: :pleading_face:
> **TA**: _[Checks the autograder: Student has commits on two days but was one day short of the threshold, and just tried to submit the day **before** the deadline. Met all other commit requirements.]_
> **TA**: ... Since we’re still a day before the deadline, I can easily waive you past the commit verification requirement.
> **Student**: So I should just resubmit tomorrow? Would the autograder let me resubmit since it says I’m blocked?
> **TA**: _[Resolves the situation]_

## Proposal
When evaluating the `git` minimum day requirement, we perform the following algorithm (notice the new steps 2, 3, and 4):
1. Do the existing check (`days_with_commits >= minimum_days_with_commits_phase_threshold`)
2. If they pass this check, then they pass the minimum days requirement
3. Count the number of full days between their most recent commit and the phase deadline (call this `early_days`)
4. If `early_days + days_with_commits >= minimum_days_with_commits_phase_threshold`, then pass the requirement.
    * Consider also displaying a notification explaining that we gave them grace on the commit penalty because they finished the project early
5. Otherwise, fail the requirement as normal

### Technical Notes
1. This will require giving some additional context to the outermost layer of the commit verification system since it currently has no idea when the assignment is due. It is already divided into layers of responsibility, and the inner layers will not need to be modified.
[...]

PR #532, opened 2025-01-30
Reduce commits out of bounds to warnings only

### LINKED ISSUE #525: Backend: Reduce "Commits Authored Before Previous Phase Hash" to warning
There is no need to block students for this. We'll just not count those commits, and warn them instead. The system already avoids counting those commits.

Image

frozenfrank:
## Overview

Reduces error messages when commits are discovered out of bounds. These commits have always been excluded, but now it will result in a warning only instead of an error message.

Resolves #525.

## Testing

- [x] Tested edge cases

## Additional Notes

It was easy to make this change because of previous improvements in #442.


frozenfrank:
Should I add a test case to verify that this condition holds going forward?

webecke:
Thats pretty neat you can request a review from a group, I'll have to do that going forward.

> Should I add a test case to verify that this condition holds going forward?

I would vote yes

19mdavenport (approved):
Looks good to me. With this do we still need the forgiveness minutes stuff?

frozenfrank:
Yes. The forgiveness minutes really account for clock-skew, while also
protecting us against students intentionally changing their clock in order
to get credit for additional commits.

frozenfrank:
Test cases added ✅

Issue #517, opened 2025-01-16
Frontend: Enforce Phase Submission after Earlier Phases Passed


frozenfrank:
## Problem

If students submit the `Quality` phase without submitting the `GitHub Repo` phase, then when they attempt to submit an actual phase, they receive this error:

> Internally failed to evaluate commit history: After processing a non-empty set of passing submissions, our latestTimestamp timestamp is null.

## Discussion

Students should be completing the GitHub Repo phase anyways. It's an easy set of points for them. Let's point in the right direction and make sure they have submitted it before continuing.

## Proposed Solution

Disable submissions on graded phases until the GitHub repo phase is submitted.

## Solution Generalization

- This proposal has been to #535

frozenfrank:
@webecke I supplied an issue description and extended the idea as we originally discussed it. What do you think of the generalization?

ZakkeryDaRebel:
Something I was thinking about since James @frozenfrank brought this to my attention. 

We already can turn off submissions to a phase and not allow them to submit and put a div in autograder/src/main/resources/frontend/src/views/HomeView.vue with c-if="isPhaseDisabled()" and id="submissionClosedWarning" saying "Submissions to this phase are currently disabled". 

We can create a new div that contains information saying "To submit to this phase, please submit to the prior phase". Or we can have a series of divs that say, "To submit to phase 0, submit to the GitHub Repository drop down option." This would be in the save area as the above div. 

To help incorporate this, in HomeView, in the div with id="submitDialoge" we need to have the submit button be disabled also if isPriorPhaseSubmitted() and the div that has the information would also need div v-if="isPriorPhaseSubmitted()"

isPriorPhaseSubmitted() would be a const written in the script section at the top of HomeView which needs to check SubmissionInfo.vue or SubmissionHistory.vue or something (I still don't fully understand the autograder) to see if the student has submitted to this phase at the very least or have passed it off successfully (no matter the grade so long as there is a grade and not a failed to compile error or similar), returning false if it can find something, or false if it can't find any history.  

Though I could work on this, I am uncertain of my skills as this uses HTML which I haven't touched in a year or two, and CSS and TypeScript which I haven't work with at all.

ZakkeryDaRebel:
@19mdavenport fixed the original issue of submitting to Quality before GitHub Repository Assignment by adding a line of code to GitHelper. 

[...]

PR #534, opened 2025-01-30
Michael's fix for the Quality phase being submitted before the GitHub…

ZakkeryDaRebel:
Michael's proposed solution to fixing issue #517 with a simple 1 line fix to stop the code from breaking.

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -314,19 +314,21 @@ public class GitHelper {
         if (passingSubmissions == null) {
             throw new GradingException("Cannot extract previous submission date before passingSubmissions are loaded.");
         }
-        if (passingSubmissions.isEmpty() || passingSubmissions.stream().noneMatch(sub -> PhaseUtils.isPhaseGraded(sub.phase()))) {
+        if (passingSubmissions.isEmpty()) {
             return MIN_COMMIT_THRESHOLD;
         }
 
         Instant latestTimestamp = null;
         String latestCommitHash = null;
         Repository repo = git.getRepository();
+        boolean hasCandidateSubmission = false;
 
         Instant effectiveSubmissionTimestamp;
         try (RevWalk revWalk = new RevWalk(repo)) {
             for (Submission submission : passingSubmissions) {
                 if (!PhaseUtils.isPhaseGraded(submission.phase())) continue;
 
+                hasCandidateSubmission = true;
                 effectiveSubmissionTimestamp = getEffectiveTimestampOfSubmission(revWalk, submission);
                 revWalk.reset(); // Resetting a `revWalk` is more effective than creating a new one
                 if (latestTimestamp == null || effectiveSubmissionTimestamp.isAfter(latestTimestamp)) {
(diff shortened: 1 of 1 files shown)


19mdavenport (commented):
Looks good to me, but I feel like I should wait for someone else to approve this

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
```diff
@@ -314,7 +314,7 @@ private CommitThreshold getMostRecentPassingSubmission(Git git, Collection<Submi
         if (passingSubmissions == null) {
             throw new GradingException("Cannot extract previous submission date before passingSubmissions are loaded.");
         }
-        if (passingSubmissions.isEmpty()) {
+        if (passingSubmissions.isEmpty() || passingSubmissions.stream().noneMatch(sub -> PhaseUtils.isPhaseGraded(sub.phase()))) {
```
frozenfrank, line 317: This is a great solution! However, I would prefer to put the solution closer to the line problem (line 340).

frozenfrank (approved):

frozenfrank:
> Looks good to me, but I feel like I should wait for someone else to approve this

@19mdavenport It's been approved. 
@ZakkeryDaRebel Let's merge this thing!

ZakkeryDaRebel:
Is this good to merge, or do we still need more people to approve? I don't know the exact requirements.

frozenfrank:
> Is this good to merge, or do we still need more people to approve? I don't know the exact requirements. 

@ZakkeryDaRebel Merge it!

ZakkeryDaRebel:
It seems like there is a Maven Test problem with this, does anyone else see it?

frozenfrank:
> It seems like there is a Maven Test problem with this, does anyone else see it?

What do you mean? It passed the automated tests, and I ran the tests locally many times.

frozenfrank:
> It seems like there is a Maven Test problem with this, does anyone else see it?

I see that the workflow failed once. I simply "reran failed tests," and [now they've succeeded](https://github.com/softwareconstruction240/autograder/actions/runs/13122200545).

I noticed during local testing that there is a small probability that the "Amended Commits" test fails if the timestamps occur incorrectly. The timestamps need to all be in the same second, but if the test takes longer and they fall on different seconds, then the test will fail.

This happens with a low probability, but I wonder if there is a way to guarantee it _never_ happens...?

ZakkeryDaRebel:
I got an email that said that the Maven tests failed, and when I updated my local main, there was originally a red checkmark next to that merge, though that has now been replaced with a green checkmark. I'm not sure what would guarantee it never failing. But is it detrimental that we merged it without it working 100% of the time?

frozenfrank:
> I got an email that said that the Maven tests failed, and when I updated my local main, there was originally a red checkmark next to that merge, though that has now been replaced with a green checkmark. I'm not sure what would guarantee it never failing. But is it detrimental that we merged it without it working 100% of the time?

@ZakkeryDaRebel You are exactly right, flaky tests are bad, and this is a prime example.

The test that caused the problem this time was not introduced in this PR. It has already been in the repo working most of the time (_this is the first time it failed, actually_).

While the particular test has been working OK so far, this is a good impetus to look at it more closely and improve it.

ZakkeryDaRebel:
So this isn't a problem with the code fix per say, but possibly an issue with the Maven Test?

frozenfrank:
> So this isn't a problem with the code fix per say, but possibly an issue with the Maven Test?

That's correct. To clarify "Maven Test" is the public name for all of the unit tests we've written in the `tests/` folder. It's effectively saying that our unit tests failed similar to the students failing the passoff tests in their project.

PR #536, opened 2025-01-31
Backend: Add Phase 3 Verifier for Location of Static Web Resources

### LINKED ISSUE #441: Backend: Add Phase 3 Verifier for Location of Static Web Resources
## Overview
Some students are able to view their static web pages just fine locally, but fail the test on the autograder because they are misplaced.

## Description
Some students accidentally mistype the name of the `resources` directory and end up serving the static web to a different URL such as `/resuorces` or otherwise. Things like this work just fine locally because the references are contained to the project, but then break when we impose the external requirements of the autograder on them.

Resolving this kind of error typically requires TA intervention since there are no debugging hints to follow to lead the student to the right answer. The goal is to provide enough information to students so that reasonable CS minor students could figure out what is wrong on their own without relying on external help.

Note that students are given clear direction in the [Getting Started](https://github.com/softwareconstruction240/softwareconstruction/blob/main/chess/3-web-api/getting-started.md) guide. We can still do more to help them recover from mistakes.

## Proposal 
Leverage the "Verifier" system already in place to double check that students have their web static files in the right location and are serving them to with the right base path.

This should be a pretty simple process, and it would only need to run in Phase 3.

When the error conditions are identified, we could provide a warning similar to the following:
> It looks like you won't pass the static files test. Are the static files located in `server/src/main/resources/web`?

Maillman:
## Overview
Resolves #441.
This pull request adds a StaticFilesVerifier to test if the web folder containing the static web files is contained in the `resources` directory for phase 3. If the web folder isn't present, the verifier returns a warning saying `It looks like you won't pass the static files test. Are the static files located in server/src/main/resources/web?`

## Testing
- [ ] Tested edge cases
- [X] Manual testing (if needed) - Tested when the resources folder is both misspelled and correctly spelled.
StaticTestVerifierWorking
StaticTestVerifierWorking2

## Additional Notes
I also looked at the ProjectStructureVerifier.java file and noticed that it checks if the `server/src/main/resources` file path exists.

https://github.com/softwareconstruction240/autograder/blob/177b77501a4dd2b089e8bf89e28bc44053139d10/src/main/java/edu/byu/cs/autograder/compile/verifers/ProjectStructureVerifier.java#L35-L37

However, in my tests, it never gives a warning saying the directory couldn't be found, even though I misspelled the `resources` folder. I debugged why this was the case and found that the AutoGrader would create a `resources` folder to hold a `db.properties` file. So the warning never occurs.

ResourcesPresent


src/main/java/edu/byu/cs/autograder/compile/verifers/StaticFilesVerifier.java
```diff
@@ -0,0 +1,24 @@
...
+public class StaticFilesVerifier implements StudentCodeVerifier {
+
+    @Override
+    public void verify(GradingContext context, StudentCodeReader reader) throws GradingException {
+        if(context.phase()==Phase.Phase3){
+            String filePath = "server/src/main/resources/web";
+            File staticFiles = new File(context.stageRepo(), filePath);
+            if(!staticFiles.exists() || !staticFiles.isDirectory()){
+                context.observer().notifyWarning("It looks like you won't pass the static files test. " +
+                        "Are the static files located in %s?".formatted(filePath));
```
19mdavenport, line 20: "Won't" seems a little strong (also technically not guaranteed to be true, but very likely).

I would also suggest putting what this detected before the problem it may cause. To someone reading this that may come off as less accusatory. 

Might I suggest: 'Could not locate folder %s. This may lead to the autograder giving different results than your local machine for the "static files" test. Please ensure the static files are in the location described in the getting started instructions and have been committed and pushed to your repository.'

19mdavenport (changes requested):
Looks generally pretty good. I have some notes about the warning message that I commented on those specific lines

19mdavenport (approved):
LGTM

frozenfrank (commented):
Does this need to be an entirely new verifier? Can you just add a line to look specifically for the `.../web/` directory which would not have the issues of the existing check?

I'm think that if there is a way to reuse the existing code, we should.

Thanks for all the research and background on the issue!

Maillman:
@frozenfrank, do you still want to give the warning to students indicating they may fail the static web tests due to the file not being in the correct place? The ProjectStructureVerifier only warns about the directory not being found.

https://github.com/softwareconstruction240/autograder/blob/177b77501a4dd2b089e8bf89e28bc44053139d10/src/main/java/edu/byu/cs/autograder/compile/verifers/ProjectStructureVerifier.java#L47-L48

frozenfrank:
> @frozenfrank, do you still want to give the warning to students indicating they may fail the static web tests due to the file not being in the correct place? The ProjectStructureVerifier only warns about the directory not being found.
> 
> 
> 
> https://github.com/softwareconstruction240/autograder/blob/177b77501a4dd2b089e8bf89e28bc44053139d10/src/main/java/edu/byu/cs/autograder/compile/verifers/ProjectStructureVerifier.java#L47-L48

@Maillman What do you think about extending the functionality of the original verifier to have the option of providing custom error messages?

src/main/java/edu/byu/cs/autograder/compile/verifers/ProjectStructureVerifier.java
```diff
@@ -14,6 +15,8 @@
  * which is checked by looking for a pom.xml file
  */
 public class ProjectStructureVerifier implements StudentCodeVerifier {
+    private static final HashMap<String, String> customMessages = new HashMap<>();
+    private static final Set<String> filePaths = new HashSet<>();
```
frozenfrank, line 19: These variables being `static` could cause problems if we were ever to try grading multiple students at once. Unless there is a clear need, since these data structures are intended to be changed, let's keep them non-static, instance variables.

src/main/java/edu/byu/cs/autograder/compile/verifers/ProjectStructureVerifier.java
```diff
@@ -27,25 +30,38 @@ private void verifyRootPom(GradingContext context) throws GradingException {
...
+                addFilePaths("server/src/main/java");
+                addFilePaths("client/src/main/java");
         }
 
         for(String filePath : filePaths) {
             File file = new File(context.stageRepo(), filePath);
             if(!file.exists() || !file.isDirectory()) {
-                context.observer().notifyWarning("Directory %s could not be found".formatted(filePath));
+                context.observer().notifyWarning("Directory %s could not be found.".formatted(filePath)
+                        + customMessages.get(filePath));
```
frozenfrank, line 64: I would make it the responsibility of this utility code to add the space, rather than require users of the function to add a leading space.

frozenfrank (changes requested):
Almost there! Just a few adjustments, and this will be a helpful addition to the AutoGrader!

src/main/java/edu/byu/cs/autograder/compile/verifers/ProjectStructureVerifier.java
```diff
@@ -60,8 +60,10 @@ private void verifyDirectoryStructure(GradingContext context) {
         for(String filePath : filePaths) {
             File file = new File(context.stageRepo(), filePath);
             if(!file.exists() || !file.isDirectory()) {
-                context.observer().notifyWarning("Directory %s could not be found.".formatted(filePath)
-                        + customMessages.get(filePath));
+                String errorMessage = "Directory %s could not be found.".formatted(filePath);
+                String customErrorMessage = customMessages.get(filePath);
+                errorMessage += (customErrorMessage.isEmpty() ? "" : " " + customErrorMessage);
```
frozenfrank, line 65:
```java
if (customErrorMessage.isEmpty()) errorMessage += " " + customErrorMessage);
```

frozenfrank (approved):
Looks great! Let's get this thing merged in!

Good job iterating on the design multiple times to arrive at a great solution, @Maillman!

PR #537, opened 2025-02-03
Fix flakey commit amend test...

frozenfrank:
## Overview

Resolves an issue that causes the `GitHelperTest:amendedCommitsCountedOnce()` test to sometimes fail based on external factors (the clock). This comes in response to this surprise discovery https://github.com/softwareconstruction240/autograder/pull/534#issuecomment-2632038772.

## Discussion

'Flakey' is a strong word because the test did nearly all of the time it ran.

Root cause argument:
* Detecting an amended commit requires an exact timestamp match. 
* Git timestamps have second precision.
* Without actually using the `commit --amend` flag, the test would create multiple commits in quick succession which is effectively the same as amending commits. 
* As long as all the commits are authored in the same second, the test works.
* However, there is a non-zero chance that the span of commits would cross the boundary of a second and cause the test to fail.
* The exact probability has to do with the execution time of the machine creating the commits.
* ∴ Therefore, probability that the test fails under valid circumstances is **non-zero**.

This change leverages a different overload of `GitHelperUtils.makeCommit()` to specify an exact Instant where the commit should be authored. We compute `now()` once, and then determine the timestamps of the other commits based on subtractions from now(). This guarantees us that the commits will *always* have the timestamps we intend for them to have.

## Testing

- [ ] Tested edge cases

## Additional Notes

This change does not fundamentally change the way the test is approached. It only guarantees that is always works properly.

### DIFF
--- a/src/test/java/edu/byu/cs/autograder/git/GitHelperTest.java
+++ b/src/test/java/edu/byu/cs/autograder/git/GitHelperTest.java
@@ -4,6 +4,8 @@ import edu.byu.cs.analytics.CommitThreshold;
 import org.eclipse.jgit.api.Git;
 import org.junit.jupiter.api.*;
 import java.io.File;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.*;
 
 class GitHelperTest {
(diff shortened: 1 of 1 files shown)


19mdavenport (approved):
That explains why it performed differently between machines and even if it was on a machine directly or inside a container. Thanks

PR #538, opened 2025-02-04
432 intelligently evaluate history without tail hash

### LINKED ISSUE #432: Backend: Intelligently Evaluate Available Data in Repo Even When A Tailhash is Missing
## Overview
In #420, we fixed an issue that caused the commit verification system to reject grading and fail to save a submission if a tail hash was missing during evaluation. Now the student will still be failed, but a submission object will be created which allows for manual intervention to apply the score for the student.

However, in all the relevant use-cases, the students did not really mean to nuke their repo. Typically, they are still setting up the project for the first time, run in to issues, and reset everything in order to fix it (after submitting the GitHub assignment). There is no need for manual intervention in these use-cases. There should be enough information for us to make a fair grading decision without manual intervention even if the tail commit is no longer available.

## Solution
Currently, we filter the commits in the `main` branch by only the ones that were not included in the previous submission (this requires a valid tail commit) _and_ then exclude other corner cases by verifying the commit timestamp. This double verification is the most robust, but we could relax the verification and still feel confident that the system is behaving as it should when a tail commit is not available.

This relaxation of requirements could be easily accomplished using existing architecture in the codebase. The `GitHelper.getMostRecentPassingSubmission()` method is responsible for identifying the latest submission commit and the latest submission timestamp and then merging the two pieces of information together into a `CommitThreshold` record. Currently, the logic tracks them together and updates both values when possible, but we could separate these concerns to update the latest timestamp, and update the latest submission hash only when it exists. The rest of the logic down the line already applies both of the filters independent of each other. This would result in a relaxation of the requirements and save students from needing a manual intervention if they blow up their repo and restart. It will still verify that the new commits were authored after the latest submission and avoid double-counting commits towards the minimum requirements.

## Potential Concerns
1. There is a strong use-case for providing this relaxation on Phase 0 (immediately following the GitHub assignment), but we may not want to have this potential loophole open for later phases where there is no demanding use-case.
    * This is not opening up an actual loophole since we are verifying author timestamps, and those author timestamps have been verified the entire time.
    * We could limit this relaxation as an exception on the Phase 0 phase. (If a student were to have this issue on a future phase, they would need a TA to approve the score.)

## Moving Forward
Feedback is welcomed. Please comment and weigh in on this issue.

frozenfrank:
## Overview

Upgrades Commit Verification internals to produce an accurate result even when student nuke their repos before Phase 1. Resolves #432.

## Details

Although I thought I would need to make some minor modifications to the algorithm, in the end, it appears that we only needed a configuration-type change in `DefaultGitVerificationStrategy`. I see that as a benefit of well designed code!

This PR does update test cases to function with the new paradigm.

## Testing

- [x] Tested edge cases

## Future Work

While working on this PR, I remembered that the `git` tests work at a low level and never test the integration of `Submission` handling with the git commits together. Future developers could add higher-level, integration-type tests which properly exercise and demonstrate correct behavior of the entire system.

## Additional Notes

At this point, I do not expect any more students to have this issue this semester, but someone will probably surprise me. At any rate, we should manually test this with a legit repo and then merge it in so we don't repeat the issues against next semester.

### DIFF
--- a/src/test/java/edu/byu/cs/autograder/git/GitHelperTest.java
+++ b/src/test/java/edu/byu/cs/autograder/git/GitHelperTest.java
@@ -261,20 +261,21 @@ class GitHelperTest {
         utils.evaluateTest("missing-tail-commit", List.of(
                 new VerificationCheckpoint(
                     repoContext -> {
-                        utils.makeCommit(repoContext, "Change 1", 24, 39, 20);
+                        utils.makeCommit(repoContext, "Change 1", 1, 9, 20);
                     },
                     utils.generalCommitVerificationResult(true, 1, 1)),
                 new VerificationCheckpoint(
                     repoContext -> {
-                        utils.makeCommit(repoContext, "Change 2", 23, 38, 10);
+                        utils.makeCommit(repoContext, "Change 2", 0, 3, 10);
 
                         // NOTE: I tried putting in an *obviously* incorrect head hash for testing, but it JGit rejected
                         // it with an InvalidObjectIdException. Apparently the ObjectIds cannot be any alphanumeric string.
-                        utils.setPrevSubmissionHeadHash("f6fbf36bd4f932177df1bc70fbd5a32da288c6d7"); // Commit doesn't exist
+                        utils.setPrevSubmissionHeadHash(null); // Commit wasn't found during processing
+                        utils.setPrevSubmissionTimestamp(Instant.now().minus(Duration.ofMinutes(30))); // Submitted between the two phases
                     },
-                    // Since the tail hash doesn't exist, it will evaluate the entire repository resulting in 2 commits on two days.
+                    // Since the tail hash doesn't exist, it will evaluate the entire repository. On the resulting in 2 commits on two days.
                     // It will be flagged as potentially incorrect and require manual intervention.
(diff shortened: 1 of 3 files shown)


Maillman (approved):
LGTM.

src/test/java/edu/byu/cs/autograder/git/GitHelperTest.java
```diff
@@ -258,23 +258,25 @@ void verifyRegularCommits() {
...
+                        utils.makeCommit(repoContext, "Change 1", 1, 9, 20);
                     },
                     utils.generalCommitVerificationResult(true, 1, 1)),
                 new VerificationCheckpoint(
                     repoContext -> {
-                        utils.makeCommit(repoContext, "Change 2", 23, 38, 10);
+                        utils.makeCommit(repoContext, "Change 2", 0, 3, 10);
 
                         // NOTE: I tried putting in an *obviously* incorrect head hash for testing, but it JGit rejected
                         // it with an InvalidObjectIdException. Apparently the ObjectIds cannot be any alphanumeric string.
```
19mdavenport, line 273: Thank you for including this note

19mdavenport (approved):
LGTM

PR #541, opened 2025-02-14
Truncate Student Error Output Data

AldenKim:
Wrote code to trim extremely long error outputs for the autograder potentially. Currently, the max characters is set to 10,000. Tested to make sure that errors would show up on the auto grader correctly


src/main/java/edu/byu/cs/autograder/test/TestHelper.java
```diff
@@ -176,4 +181,12 @@ private static String removeSparkLines(String errorOutput) {
         lines.removeIf(s -> s.matches("^\\[(main|Thread-\\d*)] INFO.*$"));
         return String.join("\n", lines);
     }
+
+    private static String trimErrorOutput(String errorOutput) {
+        errorOutput = removeSparkLines(errorOutput);
+        if (errorOutput.length() > MAX_ERROR_OUTPUT_CHARS) {
+            return errorOutput.substring(0, MAX_ERROR_OUTPUT_CHARS);
+        }
```
19mdavenport, line 189: Do you think it might be a good idea to include a message at the end stating that the output was truncated? That way the user isn't confused if it's missing something.

19mdavenport (commented):
Looks pretty good so far, thanks!

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
AldenKim: Yes, I can add a message for that before I return the truncated string. Is there an error that we can recreate to test the truncation?

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
```diff
@@ -33,6 +33,11 @@ public class TestHelper {
      */
     private static final String junitJupiterApiJarPath;
 
+    /**
+     * Constant value for trimming error outputs
+     */
+    private static final int MAX_ERROR_OUTPUT_CHARS = 10000;
```
frozenfrank, line 39: @webecke Has been working on storing configurable values in the database. This makes it easy to change dynamically. I love that we are providing a default value, but can we also make sure the number is possible to configure remotely?

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
frozenfrank: Just create a string of a really long length. On that note, can we add a test case to verify that the truncating behavior works properly? This file may not already have test cases, but we have been adding test cases to other files which you can use as an example.

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
```diff
@@ -176,4 +181,12 @@ private static String removeSparkLines(String errorOutput) {
         lines.removeIf(s -> s.matches("^\\[(main|Thread-\\d*)] INFO.*$"));
         return String.join("\n", lines);
     }
+
+    private static String trimErrorOutput(String errorOutput) {
+        errorOutput = removeSparkLines(errorOutput);
+        if (errorOutput.length() > MAX_ERROR_OUTPUT_CHARS) {
+            return errorOutput.substring(0, MAX_ERROR_OUTPUT_CHARS);
```
frozenfrank: In this case, it may be better to use a single-exit-point pattern, rather than `return`ing from multiple locations.

What do you think about this construction:

```java
        errorOutput = removeSparkLines(errorOutput);
        if (errorOutput.length() > MAX_ERROR_OUTPUT_CHARS) {
            errorOutput =  errorOutput.substring(0, MAX_ERROR_OUTPUT_CHARS);
            // TODO: Insert truncation message
        }
        return errorOutput;
```

This construct emphasizes the role of this `if` clause as **modifying** the existing string. If we wanted to add other modifying behavior in this function, that would be straightforward to chain together.

frozenfrank (changes requested):
This is great, and it does an amazing job solving the core problem.

There may be some ways we can make the overall system more adaptable in it's context in a production software environment.

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
frozenfrank: @AldenKim I think a truncation message belongs best at the **end of the string** rather than the beginning. This keeps the notification as close to the source as possible. If a student is reading at the end of the error output, the end of the output is the correct location to include this kind of message.

A simple ellipses (…) would be a start, but a more explicit message would be appropriate as well.

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
AldenKim: Sounds good. I will take a look at these.

AldenKim:
Only two things required of this pull request would be A. Writing tests to test the errorOutput truncation B. The value for Maximum characters is hardcoded, make it configurable. We were considering making new issues for these in the TA meeting.

frozenfrank:
> Only two things required of this pull request would be A. Writing tests to test the errorOutput truncation B. The value for Maximum characters is hardcoded, make it configurable. We were considering making new issues for these in the TA meeting.

Good summary

> A. Writing tests to test the errorOutput truncation

I can help make a test for this quickly. @AldenKim  LMK if I should stop and let you do it instead. 

> B. The value for Maximum characters is hardcoded, make it configurable

I would like to give it a shot right now. I expect it's as easy as just a few lines of code to create a new configurable value and read it from the config. @webecke Can we look at this together briefly?

frozenfrank:
@AldenKim I converted this to a draft so that the tests don't run on every push while this is under development. I expect we're close to merging this in.

src/main/java/edu/byu/cs/autograder/test/TestHelper.java
frozenfrank: Resolved in 3b2d468

frozenfrank (approved):
This looks great! Great work @AldenKim. This is ready to be merged.

Issue #554 has been created to represent tests. Typically, we wouldn't put off the tests like this, but we want the functionality merged in sooner. Let's finish the tests this semester.

PR #548, opened 2025-02-22
Add code coverage reporting for unit tests

19mdavenport:
## Overview
Adds code coverage reporting for unit tests, but doesn't grade it yet (wait until fall)

## Testing

- [ ] Tested edge cases

## Future Work
After enough data has been gathered and the professors make a decision about how to grade code coverage, a future TA will need to update the `UnitTestGrader`


19mdavenport:
> Referring to commit [7dd3b35](https://github.com/softwareconstruction240/autograder/commit/7dd3b3519b2812658d0261400b88d0cd3b3228ba).
> 
> Do we have to manually include the jacoco jars in our codebase? What prevents us from importing them via Maven dependency (or similar)?

Probably. We also have a junit jar and a quality checker jar in that same directory. Do we need those?

frozenfrank:
> > Referring to commit [7dd3b35](https://github.com/softwareconstruction240/autograder/commit/7dd3b3519b2812658d0261400b88d0cd3b3228ba).
> 
> > 
> 
> > Do we have to manually include the jacoco jars in our codebase? What prevents us from importing them via Maven dependency (or similar)?
> 
> 
> 
> Probably. We also have a junit jar and a quality checker jar in that same directory. Do we need those?

If there is any possible way to remove them, we should do that. Maven is the way to manage dependencies, and it works, and we should use it. 

Unless there is a serious technical reason that creates a blocker, we should not be including raw binaries in our source code. (Exceptions exist for obfuscated files for students, and for proprietary software not available on Maven; none of those apply in this case.)

frozenfrank:
@19mdavenport Yes, thank you. With that issue now understood, I'll give this another look to review the actual contents of the PR. Thanks for indulging and giving me the rest of the context on that portion of the system.

src/main/resources/frontend/src/utils/utils.ts
```diff
@@ -153,6 +154,37 @@ export const generateResultsHtmlStringFromTestNode = (node: TestNode, indent: st
...
 };
 
+const proportionToColor = (proportion: number) => {
+  /*
+  Was having a grand old time with math functions for this.
+  The point is that at 0% coverage it's max red, at 100% coverage it's max green
+      (although 255 green is really bright so 210 for now)
+  At 50% coverage it's 50% green but still about 70% red
+  */
+  const red = 255 * Math.cos((proportion * Math.PI) / 2);
```
webecke, line 164: That's wild

webecke (approved):
As far I can tell it should be good. I'm not super well versed in this part of the system though. But for now...

LGTM

src/main/resources/frontend/src/utils/utils.ts
19mdavenport: Wild in a good, bad, or neutral way?

PR #550, opened 2025-02-22
Improve test analyzer

frozenfrank:
## Overview

Fixes an issue where student test cases could be undercounted 

## Details

- Read the commit messages for more details
- This PR also cleans up some whitespace in related files

## Testing

- [ ] Tested edge cases


frozenfrank:
Simpler = better

In general, switching to a `List` instead of a `Map` would be the best solution that reduces technical debt, but this is the minimum adjustment to get the results we want. I think it's fine, but I can log an issue to perform this refactor in the future.

Funny enough, most of the code changes in this PR aren't directly related to the fix-- they're just white space adjustments to standardize the formatting of code 😃 

For future reference, the actual solution is commit 7367fa4.

PR #556, opened 2025-03-01
Fix Resubmission error

19mdavenport:
James is a silly goose

@frozenfrank Can you think of any potential unfortunate consequences of this change?

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
+++ b/src/main/java/edu/byu/cs/autograder/git/GitHelper.java
@@ -215,7 +215,8 @@ public class GitHelper {
         return new CommitVerificationResult(
                 verified, true,
                 0, 0, 0, false, originalPenaltyPct, failureMessage, null,
-                null, null, headHash, null
+                firstPassingSubmission.commitResult().minAllowedThreshold(), firstPassingSubmission.commitResult().maxAllowedThreshold(),
+                headHash, null
         ).toReport(null);
     }
 
--- a/src/main/java/edu/byu/cs/autograder/git/CommitVerificationResult.java
+++ b/src/main/java/edu/byu/cs/autograder/git/CommitVerificationResult.java
@@ -39,8 +39,8 @@ public record CommitVerificationResult(
         @Nullable Collection<String> warningMessages,
 
         // Debug info
-        Instant minAllowedThreshold,
-        Instant maxAllowedThreshold,
+        @NonNull Instant minAllowedThreshold,
+        @NonNull Instant maxAllowedThreshold,
         @NonNull String headHash,
         String tailHash
 ) {


19mdavenport:
edu.byu.cs.dataAccess.DataAccessException: Error inserting item into table submission\n\tat edu.byu.cs.dataAccess.sql.helpers.SqlReader.insertItem(SqlReader.java:128)\n\tat edu.byu.cs.dataAccess.sql.SubmissionSqlDao.insertSubmission(SubmissionSqlDao.java:89)\n\tat edu.byu.cs.autograder.Grader.run(Grader.java:106)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)\n\tat java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)\n\tat java.base/java.lang.Thread.run(Thread.java:1583)\nCaused by: edu.byu.cs.util.Serializer$SerializationException: java.lang.NullPointerException: Cannot invoke \"java.time.Instant.toString()\" because \"instant\" is null\n\tat edu.byu.cs.util.Serializer.serialize(Serializer.java:30)\n\tat edu.byu.cs.model.Submission.serializeObject(Submission.java:183)\n\tat edu.byu.cs.model.Submission.serializeCommitResult(Submission.java:179)\n\tat edu.byu.cs.dataAccess.sql.helpers.SqlReader.setValue(SqlReader.java:133)\n\tat edu.byu.cs.dataAccess.sql.helpers.SqlReader.insertItem(SqlReader.java:123)\n\t... 7 common frames omitted\nCaused by: java.lang.NullPointerException: Cannot invoke \"java.time.Instant.toString()\" because \"instant\" is null\n\tat edu.byu.cs.util.Serializer$InstantAdapter.write(Serializer.java:124)\n\tat edu.byu.cs.util.Serializer$InstantAdapter.write(Serializer.java:120)\n\tat com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:70)\n\tat com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.write(ReflectiveTypeAdapterFactory.java:196)\n\tat com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.write(ReflectiveTypeAdapterFactory.java:368)\n\tat com.google.gson.Gson.toJson(Gson.java:842)\n\tat com.google.gson.Gson.toJson(Gson.java:812)\n\tat com.google.gson.Gson.toJson(Gson.java:759)\n\tat com.google.gson.Gson.toJson(Gson.java:736)\n\tat edu.byu.cs.util.Serializer.serialize(Serializer.java:28)\n\t... 11 common frames omitted

19mdavenport:
Whenever someone resubmitted a phase they had already passed

src/main/java/edu/byu/cs/autograder/git/GitHelper.java
```diff
@@ -215,7 +215,8 @@ private CommitVerificationReport preserveOriginalVerification() throws DataAcces
         return new CommitVerificationResult(
                 verified, true,
                 0, 0, 0, false, originalPenaltyPct, failureMessage, null,
-                null, null, headHash, null
+                firstPassingSubmission.commitResult().minAllowedThreshold(), firstPassingSubmission.commitResult().maxAllowedThreshold(),
```
19mdavenport, line 218: On second thought, should the maxAllowedThreshold be `Instant.now()` instead?

frozenfrank: No. The value should be NULL. 

These debug values indicate the commits that the system evaluated in order to arrive at its result. In this context, `preserveOriginalVerification()` the system isn't evaluating any commits, and NULL is the best value to indicate this.

This particular behavior isn't new, but has been around since this subsystem was originally introduced.

Clearly, I didn't have a test to verify that this functioned properly. This issue should have been caught by a regression test.

The better solution in this case is to fix the bug in the serialization code to avoid an error when the field is null. See #558.

frozenfrank:
> Can you think of any potential unfortunate consequences of this change?

This still leaves the potential for a failure if a NULL value were to slip in the data structure some other way. The serializing code should be able to handle NULL, and this code was originally intended to be NULL.

frozenfrank:
Additionally, this code caused another error by assuming that `commitResult()` would never be NULL.

image

In this case, the error is especially painful because it also causes an error when attempting to manually approve the submission:

image

PR #557, opened 2025-03-01
fix javalin path params

19mdavenport:
I'm beginning to think Than didn't test this as much as I thought he would have

### DIFF
--- a/src/main/java/edu/byu/cs/controller/SubmissionController.java
+++ b/src/main/java/edu/byu/cs/controller/SubmissionController.java
@@ -87,14 +87,16 @@ public class SubmissionController {
     };
 
     public static final Handler submissionXGet = ctx -> {
-        String phaseString = ctx.pathParam("phase");
         Phase phase;
-
-        try {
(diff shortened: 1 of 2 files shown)


19mdavenport:
I'm going to merge this, hopefully you guys don't have issues with it.

webecke:
Dang, if only someone pointed out that PRs that wait for a long time often break things...

19mdavenport:
What would you have done differently in my situation? Scrapped javalin altogether?

frozenfrank:
> Dang, if only someone pointed out that PRs that wait for a long time often break things...

I think this was intended to be sarcastic. Dallin had warned us that things might go wrong.

However, I think we needed to press forward like happened. Issues always arise-- that's not a surprise. I'm just impressed that we had the team and the skills on hand to identify and resolve the issues so quickly!

If we want to avoid these kinds of dumb bugs (myself included), we all need to pitch in to add additional regression tests and create a culture of testing. Most of the issues that happened could have been reasonably caught with more complete testing (my bug included).

PR #562, opened 2025-03-06
Made code coverage visible only to admin

Maillman:
## Overview
Made code coverage only visible to admins (TAs) as students were getting confused whether they were getting docked by code coverage.

### DIFF
--- a/src/main/resources/frontend/package.json
+++ b/src/main/resources/frontend/package.json
@@ -10,7 +10,7 @@
     "build-only": "vite build --mode prod",
     "type-check": "vue-tsc --build --force",
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore",
-    "prettier:format": "prettier --write src/",
+    "prettier:format": "prettier --write src/ --end-of-line crlf",
     "prettier:check": "prettier --check src/"
   },
   "dependencies": {
--- a/src/main/resources/frontend/src/views/StudentView/RubricItemResultsView.vue
+++ b/src/main/resources/frontend/src/views/StudentView/RubricItemResultsView.vue
@@ -3,10 +3,12 @@ import {
   generateResultsHtmlStringFromTestNode,
   sanitizeHtml,
   generateCoverageHtmlStringFromCoverage,
+  isAdmin,
 } from "@/utils/utils";
 import type { TestResult } from "@/types/types";
 import PopUp from "@/components/PopUp.vue";
 import { ref } from "vue";
+import {useAuthStore} from "@/stores/auth";
 
 defineProps<{
   testResults?: TestResult;
@@ -30,6 +32,7 @@ const areErrorDetailsOpen = ref<boolean>(false);
   <span
     id="testResults"
     v-if="
+      isAdmin() &&
       testResults?.coverage &&
       testResults?.coverage.classAnalyses &&
       testResults?.coverage.classAnalyses.length > 0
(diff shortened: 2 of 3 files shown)


19mdavenport (approved):

src/main/resources/frontend/package.json
```diff
@@ -10,7 +10,7 @@
     "build-only": "vite build --mode prod",
     "type-check": "vue-tsc --build --force",
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore",
-    "prettier:format": "prettier --write src/",
+    "prettier:format": "prettier --write src/ --end-of-line crlf",
```
frozenfrank, line 13: Thanks for making this change. This is great!

src/main/resources/frontend/src/views/StudentView/RubricItemResultsView.vue
```diff
@@ -32,7 +31,7 @@ const areErrorDetailsOpen = ref<boolean>(false);
   <span
     id="testResults"
     v-if="
-      isAdmin() &&
+      isAdmin() && //TODO: Remove this line when we start grading students on coverage
```
frozenfrank, line 34: It looks like the code coverage was already restricted to admins only. Was something not working properly?

frozenfrank (approved):
I don't see a substantial change here, but the comments do look reasonable to me.

src/main/resources/frontend/src/views/StudentView/RubricItemResultsView.vue
19mdavenport: Were you looking at changes for all commits or just 1 commit? This line was added in this PR

frozenfrank: You're right. I didn't mean to be looking at a single commit, but I was effectively seeing only the changes in the 2nd commit.

This is a new addition. Great work!

Maillman: It wasn't, students were coming in asking about it, which was brought up in the meeting.

src/main/resources/frontend/package.json
TheDavSmasher: @frozenfrank @Maillman
Question: Why was this change necessary?
Bug: Adding this line to only format but not check creates a disparity between format and check since check will look for lf line ends but format will change all of them to crlf

Maillman: @TheDavSmasher 
As @19mdavenport said in #576:
> > why was the decision made to choose CRLF as the file ending instead of LF? Probably because it was proposed by a windows users ;)
> 
> I believe that at the time the git repository was set up to use CRLF line endings rather than LF endings and running prettier without the setting specified caused prettier to say it changed a ton of lines and files that only had differences in line endings

When I ran prettier myself without the setting specified, it changed a ton of lines and files that only had differences in line endings and I didn't want to have a bunch of line changes that visually make no difference nor did I want anyone reviewing my changes to have to go through a bunch of files with changes only to the line endings in order to find the changes that actually should be reviewed.
The choice to use CRLF was just because that was how the project was already set up. I don't think it particularly matters which line ending should be used, [although LF is the default for Prettier since v2.0.0](https://prettier.io/docs/options.html#end-of-line), but there definitely should be an explicit format, no matter which format is chosen.

TheDavSmasher: @Maillman Based on that and Michael's response within the PR, I've also added CRLF to prettier:check for consistency

PR #563, opened 2025-03-11
Javadocs - Model Module

Maillman:
## Overview
As suggested by @webecke, I have moved the Javadoc project to issue #564 and this pull request is now only concerned with the model module.


src/main/java/edu/byu/cs/autograder/compile/CompileHelper.java
```diff
@@ -28,12 +32,23 @@ public CompileHelper(GradingContext gradingContext) {
     private final Collection<StudentCodeModifier> currentModifiers =
             List.of(new PomModifier(), new PassoffJarModifier(), new TestFactoryModifier());
 
+    /**
+     * Verifies the code in the student repo, modifies its contents for grading,
+     * then compiles the code into a jar
```
19mdavenport, line 37: I think technically there's a jar for each module but i don't think that matters too much

src/main/java/edu/byu/cs/model/PrivateConfig.java
```diff
@@ -4,12 +4,32 @@
 import java.util.List;
 import java.util.Map;
 
+/**
+ * Represents the config table that contains information that only admins can see
```
webecke, line 8: I would reword this to `Represents the configuration values that only admins should see` just so that it doesn't imply there are two config tables.

src/main/java/edu/byu/cs/model/PublicConfig.java
```diff
@@ -2,17 +2,48 @@
 
 import java.util.List;
 
+/**
+ * Represents the config table that contains information that anyone can see
```
webecke, line 6: Like in the `PrivateConfig` JavaDoc, I'd recommend rewording this to `Represents configuration values that can be read by any user`

ZakkeryDaRebel (approved):
As someone that hasn't looked through the autograder all that much, I think these javadoc comments are very helpful in explaining each of the parameters and the intended purpose of each method.

Issue #564, opened 2025-03-13
Javadocs


Maillman:
## Overview
Some code documentation is lacking in the AutoGrader. In response to #444, this pull request adds Javadoc comments to many places in the code that don't already have them. I would love for any TA to look through the wording on these Javadoc comments and make sure they are worded sufficiently. If any TAs are still confused about the purpose of a class/method after reading its Javadoc comment, please let me know.

## Work List
Modules currently filled with javadoc comments:
- [x] #578 
- [x] #568
- [x] #569 
- [x] #587 
- [x] #592 
- [x] #578 
- [x] #563 
- [x] #578 
- [x] #588 
- [x] #589 
- [ ] #595

Maillman:
I'm not sure why this was moved to from the "In Progress 🔨" to "Done 🏁" and closed. There are still several modules I'm currently working under the scope of this issue.

TheDavSmasher:
@Maillman Most likely done on accident while moving some things around

PR #568, opened 2025-03-28
Javadocs - Autograder Module

Maillman:
## Overview
In response to issue #564, this pull request adds Javadoc comments for all the classes in the Autograder Module. I would appreciate as much feedback as possible about these comments and their wording!

### DIFF
--- a/src/main/java/edu/byu/cs/autograder/compile/modifiers/TestFactoryModifier.java
+++ b/src/main/java/edu/byu/cs/autograder/compile/modifiers/TestFactoryModifier.java
@@ -9,6 +9,7 @@ import java.io.File;
 
 /**
  * Modifies the <code>getMessageTime()</code> method in the test factory for phase 6
+ * so the WebSocketTests only wait 1 second
  */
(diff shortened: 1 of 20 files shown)


ZakkeryDaRebel (approved):
Everything looked good to me

src/main/java/edu/byu/cs/autograder/compile/modifiers/TestFactoryModifier.java
```diff
@@ -7,6 +7,10 @@
 
 import java.io.File;
 
+/**
+ * Modifies the <code>getMessageTime()</code> method in the test factory for phase 6
+ * so the WebSocketTests only wait 1 second
+ */
```
19mdavenport, line 13: Please mention that it modifies it from the students code so that they can still inject gson type adapters if they want to

19mdavenport (changes requested):
I see one area that could use with some more clarification, but otherwise looks great!

TheDavSmasher (approved):
These comments look great! I don't know enough to say more than that. though

PR #569, opened 2025-04-04
Javadocs - Canvas Module

Maillman:
## Overview
In response to issue #564, this pull request adds Javadoc comments for all the classes in the Canvas Module. As in previous pull requests, I would appreciate as much feedback as possible about these comments and their wording!

### DIFF
--- a/src/main/java/edu/byu/cs/canvas/model/CanvasAssignment.java
+++ b/src/main/java/edu/byu/cs/canvas/model/CanvasAssignment.java
@@ -3,12 +3,28 @@ package edu.byu.cs.canvas.model;
 import java.time.ZonedDateTime;
 import java.util.List;
 
+/**
+ * Represents a Canvas assignment
+ *
+ * @param id the id of the assignment in Canvas
+ * @param name the name of the assignment in Canvas
+ * @param due_at the date the assignment is due
+ * @param rubric the rubric for the assignment
+ */
 public record CanvasAssignment(
         Integer id,
         String name,
         ZonedDateTime due_at,
         List<CanvasRubric> rubric
 ) {
+    /** A rubric item for a canvas assignment
+     *
+     * @param id the id of the rubric item stored internally in Canvas
+     *           (generally something like '_1234')
+     * @param points the amount of possible points for the rubric
+     * @param description a description of the rubric item
+     *                    (generally something like 'Passes passoff test cases')
+     */
     public record CanvasRubric(
             String id,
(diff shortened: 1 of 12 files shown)


src/main/java/edu/byu/cs/canvas/model/CanvasAssignment.java
```diff
@@ -3,12 +3,30 @@
...
+ * @param id the id of the assignment in Canvas
+ * @param name the name of the assignment in Canvas
+ * @param dueAt the date the assignment is due
+ * @param rubric the rubric for the assignment
+ */
 public record CanvasAssignment(
         Integer id,
         String name,
-        ZonedDateTime due_at,
+        ZonedDateTime dueAt,
```
19mdavenport, line 17: Did you test the code with this change? I'm fairly sure this non-Java naming convention was deliberate as this is used for deserializing a specific kind of canvas response. https://canvas.instructure.com/doc/api/assignments.html

src/main/resources/frontend/src/components/AboutPage.vue
```diff
@@ -145,10 +145,17 @@ onUnmounted(() => {
         <AboutPagePerson
           name="Melvin Whitaker"
           url="https://github.com/Maillman"
-          tenure="Sep 2024-Dec 2025"
-          contributions="Helped clean up some backend code"
+          tenure="Apr 2024-Apr 2026"
+          contributions="Added Javadoc comments to the backend"
```
19mdavenport, line 149: If you want you could have just made your "contributions" section longer instead of replacing it.

19mdavenport (changes requested):
I see a code change that I think will break things but otherwise LGTM

src/main/java/edu/byu/cs/canvas/model/CanvasAssignment.java
Maillman: You are correct, the code change has been reverted!

19mdavenport: I'm not sure what happened but it doesn't look like that change actually got reverted

19mdavenport (changes requested):
I'm not sure what happened but it doesn't look like that change actually got reverted

Maillman:
I messed up the merge, hopefully, the code change is reverted now.

19mdavenport (approved):

PR #570, opened 2025-04-04
Fix Downloads tab functionality

19mdavenport:
## Overview
The Commit Analytics and Honor Checker were giving back empty files. This fixes that.

## Details
There was a method that was relying on the submissions to the GitHub Repository Assignment in Canvas (which we no longer require). It has been replaced with a method that just gives a set of NetID's of students in the section, and the code that was calling it has been updated to use the new method and retrieve the other info from the database.

## Testing

- [ ] Tested edge cases

### DIFF
--- a/src/main/java/edu/byu/cs/analytics/CommitAnalytics.java
+++ b/src/main/java/edu/byu/cs/analytics/CommitAnalytics.java
@@ -471,23 +471,25 @@ public class CommitAnalytics {
      *
      * @return A map section to map of netID to list of timestamp
      */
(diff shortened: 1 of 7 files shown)


frozenfrank:
Should we also introduce some kind of basic test to alert us if changes in other areas of the program break this functionality? This isn't something that is used very often, so a test is even more helpful as an alert mechanism.

Maillman (approved):
I agree with @frozenfrank that there probably should be some kind of test to make sure this functionality keeps working. I don't know if you want to make a new GitHub issue for creating those tests, or whether you want to add that in this pull request, but otherwise, the code itself LGTM!

PR #573, opened 2025-04-16
Enforce prior phase submission

### LINKED ISSUE #535: Generalize Requirements for Submitting Graded Phases in Order
## Overview

Generalize the solution from #517 to apply to each phase recursively.

## Discussion
What if we generalize this rule to disable all future phases until the previous phases are submitted?

A student should never submit **Phase 3** before **Phase 1**. Similarly, a student should never submit **Phase 5** before **Phase 3**. There is a linear order in which the assignments must be submitted. What if we disable all assignments until a passing submission is recorded for each phase defined before it in the list?

This is the linear list:
* GitHub Repo
* Phase 1
* Phase 3
* Phase 4
* Phase 5
* Phase 6

Non-graded phases like Code Quality phase are exempt from this rule, and are never disabled.

ZakkeryDaRebel:
## Overview
Resolves #535.

## Details
- Requires a passing submission in order for the prior phase
- Quality phase requires GitHub
- Currently doesn't respect Commit Verifications (If it passes, but is blocked by commits, can still proceed)
- OnMounted caches all submissions for the signed in student
- Introduced compareEnumValues
- Added some documentation 

## Testing
- [X] Tested edge cases

## Future Work
We faced significant issues with Phase discrepancies in the types.ts. Throughout the project, sometimes the Phase is a number, and sometimes it is a string. Ideally it would always be a number. `ListOfPhases()` currently has 3 usages that would need to be revised if we update this method to return numbers

Also need to edit caching because SubmissionHistory is retrieving the same data

## Screenshots

image

image

image

image

image


src/main/resources/frontend/src/components/config/LivePhaseConfigEditor.vue
```diff
@@ -17,7 +17,7 @@ type PhaseSetting = {
 const phases = ref<PhaseSetting[]>([]);
 
 onMounted(() => {
-  for (const phase of listOfPhases() as Phase[]) {
```
frozenfrank, line 20: Thanks for removing this unnecessary type assertion.

src/main/resources/frontend/src/views/HomeView.vue
```diff
@@ -57,6 +62,38 @@ const isPhaseDisabled = () => {
...
   return !useConfigStore().public.livePhases.includes(phaseName);
 };
+
+/**
+ * If this phase should be submitted after another Phase, returns that dependency.
+ * It forces GitHub to the front of the Phase assignmentOrder sequence.
+ * @param selectedPhase the intended Phase to grade.
+ * @returns the Phase that should have submissions, or null if no prior phase is required.
+ */
+const getPriorRequiredPhase = (selectedPhase: Phase): Phase | null => {
```
frozenfrank, line 72: Thanks for the clear JSDoc comments! This will be really helpful for future developers.

frozenfrank (commented):
Very clean, minimal implementation of the generalization to this problem.

I would recommend adding that caching layer to avoid the duplicate data call to the server for all of the submissions. That result could actually be relatively large since some students go crazy with their "Code Quality Code" submissions and other submissions.

frozenfrank:
@19mdavenport commented:
> I would argue that we shouldn't require a TA approval in case of git blocks, although a warning would be cool. Sometimes we have students who take the class in previous semesters and have already completed some phases and need to resubmit them.

TheDavSmasher:
@ZakkeryDaRebel just so you know, the last push didn't pass all tests and cannot be reviewed for merge

frozenfrank:
@TheDavSmasher The linter didn't pass, but the PR *can* still be code reviewed.

TheDavSmasher (approved):
I don't see anything wrong from what I can see and from surrounding code, so this could be merged.

frozenfrank:
@ZakkeryDaRebel Remember all that trouble we went through to figure out the `enum` handling? I just found an article that talks in-depth about why enums are outdated in TypeScript and proposes newer alternatives that avoid the pitfalls of enums. Specifically, the new alternatives would have avoided the pitfalls we were dealing with.

https://www.angularspace.com/breaking-the-enum-habit-why-typescript-developers-need-a-new-approach/

PR #576, opened 2025-04-22
Update linting

frozenfrank:
## Overview
The linter was updated in commit 1fc54cdc, but the files were never updated accordingly.

## Details

- Runs the linter on all the project files
- Resolves merge conflicts and also lints the changes in #573 
    - If that PR ends up not being merged soon, then the separate commit where those changes were introduced can be simply dropped from this PR.

## Testing

- [ ] Tested edge cases

## Dependencies

Currently, this PR depends on #573. Commit 1c47cd25354ed87cf45d3764c618725863cb2b19 could be dropped to remove this dependency.

## Future Work

None. I believe developers should already be in the habit of running the linter.

## Additional Notes

In commit 1fc54cdc, why was the decision made to choose CRLF as the file ending instead of LF? Probably because it was proposed by a windows users ;)


19mdavenport:
> why was the decision made to choose CRLF as the file ending instead of LF? Probably because it was proposed by a windows users ;)

I believe that at the time the git repository was set up to use CRLF line endings rather than LF endings and running prettier without the setting specified caused prettier to say it changed a ton of lines and files that only had differences in line endings

TheDavSmasher:
@frozenfrank PR Dependency has been fulfilled

TheDavSmasher:
@19mdavenport Something about the CRLF line endings change is that it was only added to prettier:format but not prettier:check so format and check will disagree on what line ending is required, making check report that prettier needs to be ran, but running prettier performs no changes to the code

19mdavenport:
@TheDavSmasher I'm not certain but you might be able to duplicate that option into the check item as well to make them in harmony with each other

TheDavSmasher:
@19mdavenport I'm not sure what you mean, but should I change check to include CRLF, remove CRLF from format, or add options for both opposites?

19mdavenport:
@TheDavSmasher I meant you might be able to add CRLF to the check. I'm fairly sure a standard was/is needed for the format and the git repo was expecting CRLF so that's what it was set to

TheDavSmasher:
Got it, will update check to include CRLF

TheDavSmasher (approved):
Having fixed prettier check, code looks great

PR #579, opened 2025-04-29
Code Tour

19mdavenport:
## Overview
Imrpoves the README by adding a brief overview of where the code for various systems are and explaining the folders/packages
Also moves a couple of classes I thought were in the wrong place

## Additional Notes
@webecke did I get all of the frontend stuff correct?
@TheDavSmasher is this what you were hoping for?


TheDavSmasher (approved):
@19mdavenport It should at least be enough to start, but a UML relationship diagram similar to what James did for Petshop would also be useful, but may be better suited for a separate PR.

Maillman (approved):
This looks great! I appreciate the detail you put into the summaries for each of the packages.

PR #584, opened 2025-05-06
Fix release workflow

TheDavSmasher:
Updated package.json dependencies, especially node to v22 to solve problem with GitHub Workflow


Maillman (approved):
As far as I can tell, this looks good.

19mdavenport (commented):
Seems reasonable, and seems like it works. Good job figuring that out.

PR #586, opened 2025-05-06
Removed Unused RegisterRequest Record

### LINKED ISSUE #581: Backend: Unused - RegisterRequest
I was going through the autograder code, adding javadoc comments, when I came across the RegisterRequest record. It was made by @pawlh, but it seems like it went unused. Was this record intended to do anything, or is it safe to remove the record altogether?

Image

Maillman:
## Overview
Resolves #581. The RegisterRequest record, which was used before integration with Canvas setup, is now completely unused and this PR removes it.


mewilker (approved):
I'm currently checked out on a pre-Spring commit and it looks like nothing used it then either, so this seems like a good change to me 👍

PR #587, opened 2025-05-07
Javadocs - Controller Module

Maillman:
## Overview
In response to #564, this pull request adds Javadoc comments to the controller module. As in previous pull requests, I would appreciate as much feedback as possible about these comments and their wording!

## Additional Notes
~On a related note, I'm not sure why @ThanGerlek called an exception "WordOfWisdomViolationException" and not something like "RepoAlreadyClaimedException".~
Hopefully, I managed to preserve the humor of the `WordOfWisdomViolationException` and still let interested developers be informed what the exception actually does.

### DIFF
--- a/src/main/java/edu/byu/cs/controller/exception/WordOfWisdomViolationException.java
+++ b/src/main/java/edu/byu/cs/controller/exception/WordOfWisdomViolationException.java
@@ -1,7 +1,7 @@
 package edu.byu.cs.controller.exception;
 
 /**
- * Throws whenever a user tries claiming a repo url already claimed
+ * Throws whenever a user tries claiming a repo url already claimed by another user
  */
 public class WordOfWisdomViolationException extends Exception {
     private static final String shortAndStout = "yo im a teapot";
--- a/src/main/java/edu/byu/cs/controller/exception/WordOfWisdomViolationException.java
+++ b/src/main/java/edu/byu/cs/controller/exception/WordOfWisdomViolationException.java
@@ -1,7 +1,15 @@
 package edu.byu.cs.controller.exception;
 
 /**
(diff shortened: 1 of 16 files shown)


frozenfrank:
@Maillman I agree that the `WordOfWisdomViolationException` is confusing and strangely undocumented.

After clarifying this myself, I was let in on the joke: this particular exception results in the HTTP 418 "I'm a teapot" response. This [official HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/418) is itself a joke, and this exception continues the joke by asserting that the "teapot" is a violation of the word of wisdom. lol.

Is there a better way to preserve the humor while also giving future developers a clue into what's going on? Probably.

ThanGerlek:
You could try adding a comment in the Exception class file or something, but if you can't find a way to do that without ruining the joke... then yeah, you can just revert it.

(I'm not mad. I'm just disappointed.)

src/main/java/edu/byu/cs/controller/exception/WordOfWisdomViolationException.java
```diff
@@ -1,5 +1,16 @@
...
+/**
+ * A more accurate name for this exception would be 'RepoAlreadyClaimedException'.
+ * This exception get thrown whenever a user tries claiming a repo url already
+ * claimed by another user.
+ * <br>
+ * This particular exception yields an HTTP response status code of
+ * "<a href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/418">
+ *     <code>418 I'm a teapot</code>
+ * </a>",
+ * and tea is clearly against the Word of Wisdom.
```
frozenfrank, line 12: 🤣🐣🙏

PR #589, opened 2025-05-15
Javadocs - Service Module

Maillman:
## Overview
In response to #564, this pull request adds Javadoc comments to the service module. As in previous pull requests, I would appreciate as much feedback as possible about these comments and their structure and wording!

### DIFF
--- a/src/main/java/edu/byu/cs/service/CasService.java
+++ b/src/main/java/edu/byu/cs/service/CasService.java
@@ -18,6 +18,11 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.Map;
 
+/**
+ * Contains service logic for the {@link edu.byu.cs.controller.CasController}. <br> View the
+ * <a href="https://calnet.berkeley.edu/calnet-technologists/cas/how-cas-works">Berkeley CAS docs</a>
+ * to understand how CAS, or <em>Central Authentication Service</em>, works, if needed
+ */
 public class CasService {
     private static final Logger LOGGER = LoggerFactory.getLogger(CasService.class);
     public static final String BYU_CAS_URL = "https://cas.byu.edu/cas";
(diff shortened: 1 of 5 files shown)


src/main/java/edu/byu/cs/service/CasService.java
```diff
@@ -18,10 +18,31 @@
 import java.net.URI;
 import java.util.Map;
 
+/**
+ * Contains service logic for the {@link edu.byu.cs.controller.CasController}. <br> View the
+ * <a href="https://calnet.berkeley.edu/calnet-technologists/cas/how-cas-works">Berkeley CAS docs</a>
```
mewilker, line 23: I'm going to have to read up on this... glad in it's in the docs!

src/main/java/edu/byu/cs/service/ConfigService.java
```diff
@@ -20,11 +20,33 @@
...
+ * Contains service logic for the {@link edu.byu.cs.controller.ConfigController}.
+ * <br><br>
+ * The {@code ConfigService} provides many of the features relating to getting, generating,
+ * and updating several configurable items. Such items include:
+ * <ul>
+ *     <li>A banner message to show users when they access the AutoGrader</li>
+ *     <li>When to shutdown the AutoGrader</li>
+ *     <li>What days the AutoGrader should regard as holidays</li>
+ *     <li>The live phases students are allowed to submit to</li>
+ *     <li>The penalties that should be applied for insufficient commits</li>
```
mewilker, line 33: Late penalties should also be mentioned here too.

src/main/java/edu/byu/cs/service/ConfigService.java
```diff
@@ -259,6 +348,13 @@ public static void updateCourseIdsUsingCanvas(User user) throws CanvasException,
         );
     }
 
+    /**
+     * Update the commit penalties in the database
```
mewilker, line 352: This function does more than git penalties, it does late day penalties too. Probably should mention that in the documentation.

mewilker (changes requested):
I think these are great! Just requesting changes so that late penalties are included in the configuration documentation (seeing how those aren't hard coded anymore).

TheDavSmasher (approved):
I agree with @mewilker request changes and have none to add myself as of now

mewilker (approved):

PR #593, opened 2025-06-02
Update Required Packages to Test

Maillman:
## Overview
This PR:
- Updated the passoffPackagesToTest for phase 0 from "passoff.chess.piece" to "passoff.chess.piecemoves".
- Renamed "unitTestPackagePaths" to "requiredTestPackagePaths" for better accuracy.


mewilker (approved):
In the future, it would probably be better if we could dynamically figure out the names of the packages required... who knows if or when they will get renamed again. I'm not sure how we would go about that though, and this seems like a good fix for now.

I definitely think the naming of the function from unitTestPackagePaths to requiredTestPackagePaths better reflects what the util does.

PR #598, opened 2025-06-17
Allow annotation dependencies

### LINKED ISSUE #596: Allow usage of org.intellij.lang.annotations dependency on projects
This is something that I think is "required" but something that would improve quality of life of anyone who desires to use these.

Case in point, IntelliJ provides certain annotations that are useful for code dev but that don't actually affect the functionality of the code (some tags include the NotNull or the one that brought this to my attention, the Language one).

However, since the autograder uses its own dependencies, it will remove this dependency. This means that students would have to remove all cases of these annotations that don't change how the code works (only how the IDE highlights/links code together) to submit to the autograder. The most useful of these annotations in my experience for this class is the Language("SQL") annotation, since it allows IntelliJ to provide autocomplete (not AI autocomplete, but standard autocomplete) when writing SQL string variables, and suggests and allows a built-in connection to the Database from IntelliJ (or, rather, it allows opening a console to query the database, so just a built-in workbench).

This shouldn't be too hard to do in my opinion but I'd like to hear some feedback after you experiment with this, or we can discuss more in person if needed. @softwareconstruction240/tas

TheDavSmasher:
## Overview
Added `org.intellij` and `org.eclipse.jgit` dependencies to submission injected pom.xml.

## Details
These dependencies used in the autograder were added to the pom.xml used solely when injected into students' projects when run through the autograder to allow the usage of annotations if the student decided to include them, incurring in any other changes to any students who didn't.

Closes #596 

## Future Work
If we feel like it's needed or desired, we could include these in the pom.xml given through the starter code so they can be used without problems if students decide to use them, but I don't think that is necessary, as it might confuse students. However, most students don't run their project through Maven, so this is a topic for discussion in a separate issue in the correct repository.

Likewise, a note in the instructions could be useful, but is out of scope of this PR and the issue it solves.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,13 +132,6 @@
             <artifactId>slf4j-api</artifactId>
             <version>2.0.13</version>
         </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-params</artifactId>
-            <version>5.10.0</version>
-            <scope>test</scope>
-        </dependency>
-
 
     </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,13 @@
             <version>6.9.0.202403050737-r</version>
         </dependency>
 
+        <!-- https://mvnrepository.com/artifact/org.jetbrains/annotations -->
+         <dependency>
+             <groupId>org.jetbrains</groupId>
+             <artifactId>annotations</artifactId>
+             <version>26.0.2</version>
+         </dependency>
+
         <!-- https://mvnrepository.com/artifact/io.javalin/javalin -->
         <dependency>
             <groupId>io.javalin</groupId>
--- a/src/main/resources/phases/pom/pom.xml
+++ b/src/main/resources/phases/pom/pom.xml
@@ -32,5 +32,18 @@
             <artifactId>junit-jupiter</artifactId>
             <version>5.9.2</version>
         </dependency>
+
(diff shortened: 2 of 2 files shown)


Maillman (commented):
Hmm... I read through the discussion on #596 and I can understand both sides of the argument there. I personally haven't encountered a student who has had to remove IntelliJ imports, so I don't know how common the occurrence is. I would like another TA who has more experience with dealing with these annotations dependencies to review, give their opinion, and approve this.

ZakkeryDaRebel:
I'm with @Mailman, I can see both sides of either adding it or not adding it. It doesn't seem like this addition will help all students, just some students that use it. But if it will help make things easier, I'm down to support this.

ZakkeryDaRebel (approved):

rodham:
If I understand correctly, the issue is that IntelliJ sometimes adds proprietary annotations that are not part of standard Java. Currently, students must remove these annotations from their code before running the auto-grader. By adding two dependencies to the pom.xml file used to run student code, students would no longer need to remove these annotations before submitting.
This change would be invisible to students, TAs, and professors with no effect on course content.
If my understanding is correct, this change seems pretty harmless and I support merging it into main. Any one disagree?

TheDavSmasher:
@softwareconstruction240/core

Maillman (approved):
Since another TA and a professor see no problem with adding this in, I will go ahead and also give my approval for this PR.

PR #599, opened 2025-06-17
Update Autograder Dependencies

TheDavSmasher:
## Overview
Updated some dependencies due to vulnerabilities and just in general potentially better usage.

Updated:
- Javalin: 6.3.0 -> 6.6.0
- MySQL: 8.3.0 -> 9.3.0
- Logback: 1.5.5 -> 1.5.18

Also removed a dependency that was duplicate in the pom.xml


Maillman (approved):
Seems like a reasonable enough change to me!

PR #602, opened 2025-06-27
Extra credit canvas

### LINKED ISSUE #591: Backend: Use Extra Credit item in Canvas
Phase 1 is currently the only assignment that has an extra credit opportunity. There is a spot in the Canvas rubric for the assignment, but at the moment the points and comments go into the Required Tests/Functionality Item. We probably want this item on the rubric to signal to the students that there is a chance for extra credit. However it's a little awkward looking that nothing gets written to it. 

Fixing this will probably require two things:
1. Adding an extra credit item to the Rubric.java model
2. Fixing the canvas integration to write to the correct part of the rubric

It may also be a good idea to write a backup where if the extra credit item can't be found, to add the extra credit to the total score (in case we get rid of the canvas item for some reason, but still want extra credit to be an available option).

TheDavSmasher:
## Overview
Removed Extra Credit calculation from Passoff Grader into its own class to handle extra credit separately for Canvas items.

Closes #591 

## Details
- Removed Extra Credit Analysis from most TestGrader logic that simply added it on top to be handled as a separate grade.
- Added EXTRA_CREDIT value to RubricType enum.
- Created ExtraCreditGrader to handle extra credit items.
- Added RubricConfig row to database (and updated markdowns).

## Testing
I coded my own repository to fulfill the extra credit tests, then ran the repository against my local autograder with an authorized Canvas Token.

image  
image
image
image
image

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   build-and-push-image:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     permissions:
       contents: read
       packages: write
@@ -25,10 +25,6 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 22 # Compatible Node.js version
-      
-      - 
-        name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
         
       -
         name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


TheDavSmasher:
I suppose theUpdating script should separate the two updates if we do want to keep the old ones, but that's also fair to consider simply removing them given when they were last updated. I'm ok with either, and I can update the markdown if we decide on keeping them or removing them.

Or I can update it to separate them and removing them be a separate issue

PR #604, opened 2025-07-02
fix connection pool catalog issues

mewilker:
## Overview
Used a single connection to set up the db and connection pool is initialized after, with a catalog set.

## Testing

- [ ] Tested edge cases


Maillman (approved):
Looks good to me...

PR #606, opened 2025-07-09
Improved Submission Dao and Dao Test Design

mewilker:
## Overview
Originally, when I made the DAO tests, I was simply focused on getting them to run successfully in the pipeline. This PR polishes some of the things that I left out of the original tests, and I remembered to run code coverage at the end too...

According to IntelliJ, Code coverage on SubmissionMemoryDao is now 100% method, 97% line, and 91% branch. Code coverage on SubmissionSqlDao is 100% method, 96% line, and 54% branch. Most of the uncovered methods some from SQL exceptions that aren't thrown. Others come manuallyApproveSubmission when there are multiple targeted submissions. The likelihood of this in the wild to me seems very low, since it already identifies by netid, hash, and phase, but if we deem that we should cover those branches I will.

## Details

- Tests now run both the memory implementations and the sql implmentations. There is a package called `base` where the tests live that call the given dao holds the test. The `sql` and `memory` packages hold methods to provide the daos for the tests, as well as the method to clear the tables
- While working on this, I found and fixed some bugs with the Submission memory dao. In instances where the implementation disagreed on how to do something, I typically defaulted to the sql dao's method.
- Removed the line that could cause timestamp collisions, because the memory dao didn't sort it's list of returned times and I didn't want to change that.
- Added conditions and tests to improve code coverage

## Testing

- [ ] Tested edge cases

## Future Work
I'd still like to somehow generalize the clear function in the SqlTestDaoUitls, but I have some concerns about foreign keys so I'll probably address that with the UserDaoTests.


Maillman (approved):
LGTM!

PR #614, opened 2025-09-18
#539 More Dao Tests

### LINKED ISSUE #539: Create SQL DAO Tests
## Overview

There are currently no tests to verify that SQL DAO's do their job properly. I'm currently changing some stuff, and I want assurance that it's working properly.

## Discussion

Specifically:
* `SubissionSqlDao`

mewilker:
## Overview

This PR adds unit tests for the remaining DAOs that do not have tests yet. Resolves #539. There was some clean-up as well that came as I faced choices on how to handle certain problems. Most of the Memory DAOs were changed to closer match SQL implementations.

With this PR, code coverage on the DataAccess Package results in 95% class (DaoService is not covered by these tests) 88% Method, 89% Line and 55% Branch (From what I can tell, most of the uncovered branches are SQLExceptions being thrown)

## Details

- Added and changed Dao tests, namely:
  - SubmissionDao
    - Now uses some TestUtils
    - Submission deepEquals function added after realizing that the Submission equals function is overwritten
  - UserDao
    - Memory Dao changed to match SQL behavior
  - QueueDao
    - removed unused pop method
    - return all in queue now returns users in order sorted by timestamp (both implementations)
  - RubricConfigDao
    - added null guard on getPhaseTotalPossiblePoints default implementation
    - addRubricConfig now adds the rubric ID when passed a rubric config item in the SQL implementation
    - Memory Dao changed to match SQL behavior
  - ConfigurationDao
    -  Memory Dao changed to match SQL behavior
  - RepoUpdateDao
    - Memory Dao changed to match SQL behavior
- Removed Unused Commit Phase Enum 
- Added Quality Rubric to Phase Utils

## Testing

- [x] Tested edge cases

## Future Work
Once this PR is merged into the codebase, it would probably be a good time to do some clean-up on the DAO files. The tests should give some confidence that any future changes made wouldn't affect the expected behavior.

## Additional Notes
I'm releasing this PR to a draft version because I'm still not thrilled with the performance of the tests: about 12 seconds. While I know it would have been much longer without the Connection Pool added when the Submission Tests were made, I'd still like to see if there's any small optimizations I can make to the code.  The things I plan to look into are:
- [x] Creating one connection pool for all the test files 
- [ ] (Not planned -- all tables have a primary key which means optimizing this would probably be a small gain if any)Tables that need to be indexed (and if that would require some alterations the tables)
- [ ] If there should be less dummy data generated in certain tests

If there are other ideas about how to improve the test's performance, I'd be happy to hear them.


src/main/java/edu/byu/cs/dataAccess/memory/UserMemoryDao.java
```diff
@@ -26,7 +27,7 @@ public User getUser(String netId) {
     @Override
     public void setFirstName(String netId, String firstName) {
         if (!users.containsKey(netId))
-            throw new IllegalArgumentException("User does not exist");
+            return;
```
ZakkeryDaRebel, line 30: Is there a specific reason why you replaced all these exception throwing with just a return?

mewilker: Yeah, they match the behavior of the SQL DAOs better, because if you edit something in the SQL that doesn't exist it doesn't throw an error.

ZakkeryDaRebel (approved):
Wow, that is one big pull request, but I say that I'm impressed with all the tests that you've written in here. If it all works and runs fine, then I approve it.

src/main/java/edu/byu/cs/dataAccess/memory/ConfigurationMemoryDao.java
```diff
@@ -10,12 +12,31 @@ public class ConfigurationMemoryDao implements ConfigurationDao {
...
 
     @Override
-    public <T> void setConfiguration(Configuration key, T value, Class<T> type) {
+    public <T> void setConfiguration(Configuration key, T value, Class<T> type) throws DataAccessException {
+        if (key == null || type == null){
+            throw new DataAccessException("Cannot set a configuration with a null key or type");
+        }
+        switch(type.getName()){
+            case  "java.lang.Boolean", "java.lang.String", "java.lang.Integer", "java.time.Instant",
+                  "java.lang.Float" -> configuration.put(key, value);
```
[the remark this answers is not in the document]
mewilker: I think that line 24 was leftover from the original code... the original memory dao didn't type check. I'll go ahead and remove line 24. Thanks for noticing!

PR #630, opened 2026-01-28
HOTFIX: Incorrect "You are not in queue" message

mewilker:
## Overview

This is a hotfix for the error message that says "You are not in the queue" even though users are in fact in the queue.

The root of the problem was the change in my last PR of the cookie to `HttpOnly`. These queue updates rely on websocket messages that were intiated by sending the user's auth token. In response, they got grading updates. Now that the token is inaccessable, the backend can't get the user.

This was fixed by reading the cookie instead of the websocket message.

### Before
image

image

### After
image

image

## Testing

- [ ] Tested edge cases

## Future Work

There's a good chance this will cause an error for Test Student, but I'd rather everyone else have the correct display. There's probably also some code cleanup that needs to happen, but I will address this later.


TheDavSmasher (approved):
Funny

PR #637, opened 2026-05-04
Update lower_snake_case variables from canvas to lowerCamelCase

camchamb:
## Overview

Using GSON's serializedName decorator,  lower_snake_case variables were changed to lowerCamelCase. This increases consistency across the code base while still serializing correctly from Canvas. 

## Testing

- Passed the unit tests on my local machine. 
(Some were failing, but also failed when I ran the tests on the main branch, so it's likely my auto grader setup.)

## Future Work

I'm fairly sure that I refactored all of the lower_snake_case variables from my search, but there may be more.

### DIFF
--- a/src/main/java/edu/byu/cs/canvas/CanvasIntegrationImpl.java
+++ b/src/main/java/edu/byu/cs/canvas/CanvasIntegrationImpl.java
@@ -79,7 +79,7 @@ public class CanvasIntegrationImpl implements CanvasIntegration {
                     "/courses/" + getCourseNumber() + "/sections/" + sectionID + "?include[]=students",
                     CanvasSection.class)
                 .body().students().stream()
-                .map(CanvasSection.CanvasSectionStudent::login_id)
+                .map(CanvasSection.CanvasSectionStudent::loginId)
                 .collect(Collectors.toCollection(HashSet::new));
     }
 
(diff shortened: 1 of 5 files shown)


src/main/java/edu/byu/cs/canvas/CanvasIntegrationImpl.java
```diff
@@ -167,7 +167,7 @@ private String buildRubricSubmissionQueryString(CanvasRubricAssessment assessmen
     public CanvasSubmission getSubmission(int userId, int assignmentNum) throws CanvasException {
         return makeCanvasRequest(
                 "GET",
-                "/courses/" + getCourseNumber() + "/assignments/" + assignmentNum + "/submissions/" + userId + "?include[]=rubric_assessment",
+                "/courses/" + getCourseNumber() + "/assignments/" + assignmentNum + "/submissions/" + userId + "?include[]=rubricAssessment",
```
mewilker: With this being a URL path, I'm don't think we can change this instance to lower camel case.

src/main/java/edu/byu/cs/canvas/CanvasIntegrationImpl.java
```diff
@@ -137,10 +137,10 @@ public void submitGrade(int userId, int assignmentNum, CanvasRubricAssessment as
     private String buildRubricSubmissionQueryString(CanvasRubricAssessment assessment, String assignmentComment) {
         StringBuilder queryStringBuilder = new StringBuilder();
         for(Map.Entry<String, CanvasRubricItem> entry : assessment.items().entrySet()) {
-            queryStringBuilder.append("&rubric_assessment[").append(entry.getKey()).append("][points]=")
+            queryStringBuilder.append("&rubricAssessment[").append(entry.getKey()).append("][points]=")
```
mewilker: This also seems like part of the Canvas API, and something we can't change.

src/main/java/edu/byu/cs/canvas/CanvasIntegrationImpl.java
```diff
@@ -137,10 +137,10 @@ public void submitGrade(int userId, int assignmentNum, CanvasRubricAssessment as
...
                     .append(entry.getValue().points());
             if(entry.getValue().comments() != null) {
-                queryStringBuilder.append("&rubric_assessment[").append(entry.getKey()).append("][comments]=")
+                queryStringBuilder.append("&rubricAssessment[").append(entry.getKey()).append("][comments]=")
```
mewilker: This also seems like part of the Canvas API, and something we can't change.

mewilker (changes requested):
Nice work! I think the refactor may have gone a bit too far... It looks like it edited some strings that are paths, and could potentially break the canvas integration. I would look through your changes again and make sure that you didn't change any Canvas API calls other than the ones I found. Once that's done, this should be good to go!

camchamb:
Thanks, I should have looked more closely. All the current changes should now only apply to the records and accessing them.

mewilker (approved):
Nice! Tested it to make sure Canvas still works and it does :) good job!

PR #639, opened 2026-05-04
634 honorchecker zip identification

### LINKED ISSUE #634: Honorchecker Zip Identification
Dr. Wilkerson has found it difficult to tell who's files exactly are getting flaged when he runs code through the honor checker. Currently, each student's repository is zipped into an individual folder, and then all the students are zipped into one section file. When a file is flagged for suspicious behavior, Dr. Wilkerson says that he has to track down the student based off the naming convention of the file.

He would like to see a json file at the root level of each student's zip file that identifies the student, so that it's easier to tell who the infraction belonged to. This json should contain things like first name, last name, and some sort of school identifier. There may be other attributes that would be useful to track. He wants it to be somewhat generic as other universities will end up using the honor checker.

The PR that resolves this should tag @jerodw as a reviewer.

camchamb:
Resolves #634

## Overview

I added a student_info.json inside each student's zip folder. It has the following information in it:

netId
canvasUserId
firstName
lastName
repoUrl
role: ADMIN/STUDENT
school

## Testing

- [ ] Tested edge cases
- [x] Downloading the zip file took the same time with and without the change locally. 

## Additional Notes

Would love others to test and verify as well.

### DIFF
--- a/src/main/java/edu/byu/cs/honorChecker/HonorCheckerCompiler.java
+++ b/src/main/java/edu/byu/cs/honorChecker/HonorCheckerCompiler.java
@@ -16,7 +16,6 @@ import java.io.File;
 import java.io.FileWriter;
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Map;
 import java.util.Optional;
 import java.util.function.Consumer;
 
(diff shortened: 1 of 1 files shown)


mewilker (approved):
Worked for me! I'm not sure how long this takes on the autograder normally, but it took my machine a few minutes. Would we want to timestamp the code in the json file @jerodw?

PR #640, opened 2026-05-05
Overall flow doc

### LINKED ISSUE #445: Create SequenceDiagram of Overall Grading Flow
Use GitHub [`mermaid`](https://mermaid.js.org/#/) to document the way the AutoGrader system behaves when a student requests grading for a particular system. This is part of #444.

Make sure to cover important systems such as:
* Original Request
* Server & Controllers
* Grader
* Verifiers
* Graders
* Commit Verification
* Late Penalty Calculations
* Saving Results to Gradebook & Database
* Common error paths
    * Commits not verified
    * Not all tests pass
    * Code cannot compile

TwoLettuce:
## Overview
RESOLVES: #445 
Created a document with several Sequence Diagrams to document overall grading flow. It is meant specifically new TAs that are not familiar with the Autograder, hence the sequence diagram that closely resembles the Chess Phase 2 sequence diagram to indicate familiar territory.

## Future Work
Class Diagrams and potentially more sequence diagrams for more areas of the project that weren't covered very specifically. Fixing any problems or programming principle violations exposed by the diagram.

If you have any suggestions for making things more clear or adding/removing from the diagrams, feel free to comment!


docs/overall-grading-flow.md
```diff
@@ -0,0 +1,219 @@
...
+    Service->>Service:getMostRecentSubmission()
+    Service->>Service:assertHasNewCommits()
+    
+    Service->>dao:addToQueue(netid, phase)
+    dao->>db:add submission to queue
+    Service->>Service:submit(Grader)
+```
+
+### Grading flow diagram
+The following diagram expands on the `executeGrader()` step from the previous diagram. The diagram illustrates the execution of a single Grader from start to completion.
```
camchamb: I didn't see an executeGrader() call in the previous diagram.

TwoLettuce: I went ahead and fixed the language to reflect the diagram! thanks!

docs/overall-grading-flow.md
```diff
@@ -0,0 +1,219 @@
...
+Any changes that impact the flow of the diagrams should be accompanied by corresponding updates to ensure the diagrams remain relevant.
+
+The autograder's interaction with BYU's authentication service (OAuth 2.0) is described [here](https://developer.byu.edu/data/api-usage/learn-about-oauth-2-0). 
+You will need to log in to your BYU student account to access the documentation.
+It will not be covered in the diagram because it happens before the student is granted access to the autograder and does not happen as part of the grading flow.
+
+### Autograder Submission Flow
+The following diagram provides a high-level overview, similar to the sequence diagram for phase 2 of the chess project. It traces a single submission from request to completion.
+The diagram is simplified as follows: 
+- Does not differentiate between the different Service classes or DAOs but refers to them collectively as `Service` and `DAO` respectively.
```
mewilker, line 13: I think it would be helpful to have something to distinguish which DAO/Service you are calling. Specifically the DAOs. I don't think adding an actor for each DAO is the move though, it would just clutter up your beautiful diagam. Maybe you could do something like `tablename.function()` for example `queue.remove(netid)` or `user.getCanvasUserID`.

docs/overall-grading-flow.md
```diff
@@ -0,0 +1,219 @@
...
+    git-->>Service:newest commit
+    Service->>dao:getSubmissionsForPhase(netid, phase)
+    dao->>db:get student's previous graded submissions
+    dao-->>Service:Collection<Submission>
+    Service->>Service:getMostRecentSubmission()
+    Service->>Service:assertHasNewCommits()
+    
+    Service->>dao:addToQueue(netid, phase)
+    dao->>db:add submission to queue
+    Service->>Service:submit(Grader)
```
mewilker: Do you know when the response is sent back to the user? Is there a reason you chose not to model that?

docs/overall-grading-flow.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+### Grading flow diagram
+The following diagram expands on the `submit(Grader)` call at the end of the previous diagram. The diagram illustrates the execution of a single Grader from start to completion.
+Multiple Graders are managed by an [Executor Service](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ExecutorService.html) with a threadpool size of 1. 
+Upon completion, a `Submission` is created and uploaded to the database. The diagram is simplified as follows:
+- `PreviousPhasePassoffTestGrader` is abstracted and is represented by Grader.
+- Canvas integration is abstracted and modeled [here](#canvas-integration-diagram). All interaction with canvas happens immediately before `Scorer` returns the `Submission` to `Grader`
```
mewilker: I think it would be nice if you hand a line in the diagram for this, similar to what you did above with Github. If mermaid supports colors you could color the arrow and then say something like "The details of the red arrow are coverned in the third diagram" or something like that. If it doesn't or is too difficult don't worry about it.

mewilker (commented):
I like this diagram! If I had started developing with this diagram, I think I would have understood the autograder faster. I think this will be a great introduction to the autograder for future TAs.

docs/overall-grading-flow.md
TwoLettuce: Valid. I do think it is worth showing in the diagram so I'll go ahead and fix that.

docs/overall-grading-flow.md
TwoLettuce: I chose not to model it because it happens asynchronously as Grader is executed, though I'll include the response as though it was handled synchronously. That's also a main reason I put the execution of Grader into a separate diagram.

mewilker (approved):
Nicely done! I think the diagram shows some places were we could look into optimizing as well if we wanted to, such as the multiple rubric config calls.

bsharplydian (commented):
see previous comment

PR #642, opened 2026-05-20
594 frontend autograder fails to load submission details

### LINKED ISSUE #594: frontend: autograder fails to load submission details
Every once and a while, The autograder loads up like this: 

Image

In this state, clicking things loads empty boxes, and student search doesn't work. You can click the submission refresh button, and it fixes some of the problem:

Image

All data is loaded, and you can click on the blank boxes to see it. It doesn't fix the student search bar. Refreshing the page takes the whole page back to normal, so this isn't a high priority bug.

I don't know what's causing it, but I've never noticed this particularly before Spring 2025. I've been able to replicate the bug in production by closing the tab a few times and reopening it. It doesn't always happen, so you may need to try it a few times. I'm not sure if it's repeatable in a development environment.

camchamb:
## Overview

Resolves #594

The autograder submission details would sometimes not load. I believe this was because of a race condition. The table would start loading before the data was fully loaded. 

## Details

- I first added error handling if the cell doesn't have the information. In a failed case, it will just be an empty cell. 
- I changed the row update to happen the recommended way from the ag-grid-community website. Once the grid's row values are all updated, it calls the onGridReady function. Updating this way, I never ran into a problem or got the fail case in the cell renderings. 

## Testing

- [ ] Tested edge cases

I was able to test it the same way that Melvin showed in the issue. After the changes, I never ran into issues, but I would love people to check it on their own machines. Testing if the cells can be clicked would also be useful. 

## Future Work

I don't think other tables have this problem, but we could update those in the future as well.

### DIFF
--- a/src/main/resources/frontend/src/utils/tableUtils.ts
+++ b/src/main/resources/frontend/src/utils/tableUtils.ts
@@ -12,8 +12,8 @@ import {
 export const renderPhaseCell = (params: ValueGetterParams): string => {
   const phase = params.data.phase;
   if (typeof phase !== "string") {
-    console.error("Failed to get phase")
-    return ""
+    console.error("Failed to get phase");
+    return "";
   }
   return params.data.phase.toLowerCase().replace("phase", "");
 };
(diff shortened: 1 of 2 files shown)


ZakkeryDaRebel (approved):
Looks good to me. I ran this on my local machine, and tried multiple times to get the problem to occur, but it never happened, it always loaded in the information like it should. Great job!

mewilker (approved):
I also could not replicate the bug anymore!

PR #643, opened 2026-05-22
Code Coverage

### LINKED ISSUE #398: Backend: Add some verification of student test cases
Many students pass with test cases that are not what we are looking for such as
- assertTrue(true);
- writing integration tests similar to StandardAPITests instead of unit tests (detectable if they import from the "passoff" package in the passoff-dependencies.jar)

I think a bit of verification that they are not doing anything prohibited before grading their test case code would go a long way

Not sure if this should be a StudentCodeVerifier, a bit of code inside the UnitTestGrader might be better

mewilker:
## Overview
Resolves #398. NOTE: this PR **should not be merged** until all other instruction has been updated and we are ready to start grading students on this requirement.

This PR changes how unit tests are graded, rather than counting numbers of tests, it evaluates the coverage on an area of source code that is relevant to the phase. The formula for grading is (student coverage / expected coverage). The idea is that unit tests are now graded on quality not quantity.

## Details

- Changed unit test grader to use coverage to caluclate scores
- Added a verifier to ensure student code is in the correct spots to collect coverage, sends a warning if not
- Penalty config now takes in configurable code coverage requirements
- Rubric config sql statements have been changed to outline current expected coverage requirements (changing tables is probably not necessary for developers)

## Testing

- [x] Tested edge cases

## Dependencies

The softwareconstruction repository with the instruction should be changed at the same time this goes into effect.

## Future Work

Dr. Wilkerson has stated that we still may want to scan the code for bogus test assertions.

## Additional Notes
Big shout out to @ZakkeryDaRebel for taking responsibility over this for a short while and @19mdavenport for writing the code that actually collects the coverage.


src/main/java/edu/byu/cs/autograder/test/UnitTestGrader.java
```diff
@@ -45,25 +71,78 @@ protected float getScore(TestOutput testOutput) throws GradingException {
...
-        if (totalTests < minTests) return (float) testResults.getNumTestsPassed() / minTests;
+        float coveragePercent = getCoveragePercent(testOutput.coverage());
+        if (Float.isNaN(coveragePercent)){
+            return 0;
+        }
 
-        return testResults.getNumTestsPassed() / totalTests;
+        if (coveragePercent > extraCreditPercent) {
+            return 1.05F;
+        }
```
19mdavenport: Might you want the extra credit to be configurable from the frontend as well, rather than hardcoding it?

src/main/java/edu/byu/cs/autograder/test/UnitTestGrader.java
```diff
@@ -45,25 +71,78 @@ protected float getScore(TestOutput testOutput) throws GradingException {
...
+        if (coveragePercent > targetPercent){
+            return 1;
+        }
```
19mdavenport: Would you want these to be `>=` rather than `>`?

19mdavenport (commented):
I don't know if my feedback is wanted, but I noticed a couple of things that may be helpful.

In addition to the other two comments on lines, I noticed in several places you added the coverage config stuff to the "penalty" section in the config. I understand you may not want to add a whole new config endpoint, request, and other things, but do the code coverage settings qualify as penalties?

src/main/java/edu/byu/cs/autograder/test/UnitTestGrader.java
mewilker: extraCreditPercent is configurable! If it can't grab it from the dao though it defaults to 90%

src/main/java/edu/byu/cs/autograder/test/UnitTestGrader.java
mewilker: Yes thank you

src/main/java/edu/byu/cs/autograder/test/UnitTestGrader.java
mewilker: Oh wait did you mean the score?

19mdavenport: I did

src/main/java/edu/byu/cs/autograder/test/UnitTestGrader.java
```diff
@@ -45,25 +71,78 @@ protected float getScore(TestOutput testOutput) throws GradingException {
...
 
-        if (totalTests < minTests) return (float) testResults.getNumTestsPassed() / minTests;
+        float coveragePercent = getCoveragePercent(testOutput.coverage());
+        if (Float.isNaN(coveragePercent)){
+            return 0;
+        }
 
-        return testResults.getNumTestsPassed() / totalTests;
+        if (coveragePercent >= extraCreditPercent) {
+            return 1.05F;
```
ZakkeryDaRebel, line 84: Did we want to use the Extra Credit Rubric set up by DavSmasher instead of returning 1.05, or is it just handled in this one spot?

ZakkeryDaRebel (commented):
Besides my question about the Extra Credit portion, my version of the autograder did not work as expected. 

I submitted my Phase 3 code to the autograder, and with line got 80.2%, and with branch I got 77%. This gave me extra credit version of the grade (26.25/25). Then when I went into the config and changed the percentages to be 80% for full points and 90% for extra credit, my 80.2% line coverage instead gave me 22.38/25, or about 90% of the points instead of full points. 

I'll take a look at the code if there aren't any students coming in about what is causing this issue, but thought I would point out the problem and see if anyone else is experiencing issues. 

EDIT: I realized that I had accidentally changed both to 90%, so that's why I only got 90% of the points that time around.

ZakkeryDaRebel (approved):

PR #644, opened 2026-05-22
Refactor Authstore in Frontend

### LINKED ISSUE #631: Fix HttpOnly Cookie fallout
As part of PR #628, I made it so that the auth token placed in cookies were `HttpOnly`. This prevents Cross-site scripting attacks. However, this means that the frontend is unable to access the token, causing some small issues when the PR went live. This issue entails:

1. Figuring out if there are more features affected by changing the token cookie to `HttpOnly`
2. Fixing said issues.

I suspect that the Test Student may need to be tweaked. There may be other things I find that need to be changed.

mewilker:
## Overview
Resolves #631

When the authentication system switched, I changed the cookies containing the authentication tokens to be http only. This broke some features at the time that used websocket. I made a quick fix, but at the time I was not certain how much else relied on the cookie not being http only.

After looking through the frontend code a little more thoroughly, I've concluded that the code mostly relies on the meGet endpoint call. There is little to no use of the actual token cookie. The websocket seemed to be the only place that was writing the token in manually, and websocket messages are accompanied by their cookies, so this was unecessary. This change was shortly after the authentication was changed.

To avoid future uses of the cookie in the frontend, I've changed the AuthStore into a UserStore. This overall change prevents cross site scripting and is more secure for everyone.

## Details

- Changed test student to also use an http only cookie
- Removed token references from AuthStore
- Renamed AuthStore to UserStore 
- Removed unused Websocket variables and methods
 
## Testing

- [ ] Tested edge cases

### DIFF
--- a/src/main/java/edu/byu/cs/controller/AdminController.java
+++ b/src/main/java/edu/byu/cs/controller/AdminController.java
@@ -4,6 +4,8 @@ import edu.byu.cs.canvas.model.CanvasSection;
 import edu.byu.cs.controller.exception.ResourceNotFoundException;
 import edu.byu.cs.model.User;
 import edu.byu.cs.service.AdminService;
+import edu.byu.cs.service.AuthenticationService;
+import io.javalin.http.Cookie;
 import io.javalin.http.Handler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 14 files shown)


camchamb (approved):
Worked great on my computer.

Issue #636, opened 2026-04-30
Get Course Number Dynamically from Canvas Tests


bsharplydian:
Right now the course number is hardcoded on line 40 of CanvasIntegrationImplIT.java.
We need to modify the tests so they query the mysql database to get the current course number, so it doesn't need to be maintained each semester.

bsharplydian:
Discussed with Makenna; the bigger problem is that I didn't have access to the old course number with my API token. If we use Canvas API token for the Autograder user, it will be able to access all past courses.
However, I still think it would be good practice to keep the course ID current, in the case that an old course gets deleted/modified. I'll do this by getting all the courses from the Autograder user, then choosing the most recent one. This will remove the need to access MySQL from the tests.

PR #651, opened 2026-06-01
Canvas Integration Tests (at long last)

### LINKED ISSUE #65: backend: create tests for Canvas integration
I suspect that one of the most brittle parts of this project will be the Canvas Integration. For every Canvas endpoint that we are using, we should create a test to verify it's functionality

bsharplydian:
## Overview
Added further Canvas integration tests, with several dependencies. Resolves #65. Resolves #649.

## Details
- wrote tests for each of the CanvasIntegrationImpl functions.
- coverage: 100% Class, 96% Method, 90% Line, and 60% Branch.
- added default values for RubricConfigDao
- thoroughly investigated the possibility of getting course information at runtime instead of using hardcoded values (I think it's possible and I opened an issue for it)

## Testing
- [ ] Tested edge cases

## Future Work

- Find a way to get the course number dynamically, as described in #636.
- Use the default Rubric Config values for MySQL to make setting up the autograder environment easier, as described in #650.


mewilker (approved):
I believe these tests are good. We still need to integrate this into the pipeline, as well as address the comments mentioned above, but I know @bsharplydian is itching to be done with this work.

src/test/java/integration/CanvasIntegrationImplIT.java
```diff
@@ -35,16 +40,168 @@ public static void setUp() {
...
+    public void getNetIdsFromSection() {
+        Collection<String> netids = null;
+        try {
+            int sectionID = retriever.getSectionIDFromCanvas();
+            netids = canvasIntegration.getAllStudentNetIdsBySection(sectionID);
+        } catch (CanvasException e) {
+            LOGGER.error("Could not get Net IDs from current section: {}", e.getMessage());
+            fail("Exception thrown: ", e);
+        }
+        assertNotNull(netids);
```
ZakkeryDaRebel, line 88: In addition to checking to see if netids is not null, should we also check to make sure it has a list of greater than 1 student? This way we are ensuring that it is grabbing multiple net IDs?

ZakkeryDaRebel (approved):
It all looks good, I just had 1 comment on one of the tests to ensure it is returning multiple like the test states. But it isn't all that required. Other than that, everything looked good.

PR #653, opened 2026-06-05
Log Code Quality issues

mewilker:
## Overview
Resolves #648 

A student came to me with an issue with their repository.

image

This is fine... as long as the TAs know what's going wrong. In this particular instance, it was a class outside the compile scope that was missing the curly brace to finish the class. The compiler hadn't built the class because it was outside the project, but checkstyle still ran on the file and threw an exception because it wasn't valid Java.

I had to guess this. In an alternate reality, we would have spent hours debugging it.

This PR puts the error stream from the checkstyle in the result notes of this error. This gives the TAs, and maybe an astute student a place to start debugging the error.

image

image

## Testing

- [ ] Tested edge cases


ZakkeryDaRebel (approved):
This looks great! I agree that it might have taken me a long time to find the problem, so adding the details of the error stack trace will be very helpful for debugging purposes, especially since the old version has no details or any information besides it failed to compile.

PR #654, opened 2026-06-12
update passoff-dependencies.jar

mewilker:
## Overview
Updated to the latest pass-off dependencies.jar to prepare to be in sync with next term's starter code.


camchamb (approved):
Tested it on my computer, and it works great. It passed all the test and gave better errors when I wasn't passing them.

PR #656, opened 2026-06-23
Update to Java 23

mewilker:
## Overview
Resolves #645 kinda.

Originally I set out to upgrade the autograder to version 25, so that we could support the latest LTS of java possible. The current chess project is set up to run JDK 23, and the current version of the autograder is version 21. Recently, we have seen quite a few students using unamed variables, which was introduced in Java 22. They then push their code to the autograder and it fails to compile.

It turned out that Checkstyle did not support Java 25 language features. This meant that although the autograder could upgrade to Java 25 and run projects using features just fine, it could not run code quality with these features. Checkstyle did support unnamed variables, and claims to be supported by Java 23. Thus, this pull request updates the autograder to that version. This should allow students to use unnamed variables in their code, consistent with the JDK we give them to start the project.

The Dockerfile was updated to use version 25. This is because as far as I could see, it only has builds for LTS versions. Thus, the best version to use was 25.

## Details

- devcontainer now uses version 23
- tests workflows now use version 23
- dockerfile now builds with java 25
- autograder maven pom is set to use 23
- student injected poms are set to use 23
- root gitignore includes node_modules (see commit note)
- updated Mockito dependency

## Testing

- [ ] Tested edge cases

## Future Work

Should checkstyle be upgraded to support Java 25, we could make the final leap and upgrade to fully to Java 25. A feature I used to test this was placing code before the super call in a constructor. We would first need to upgrade the Checkstyle repository and then update this one.


TwoLettuce (approved):
Looks like it probably works!

PR #661, opened 2026-07-07
fix bump to incompatible maven verision

mewilker:
## Overview

The following error was given when anyone tried to compile in the autograder for version 1.2.4:
image

I was unable to replicate it on my machine, but it seems the issue was the release used Maven 4 instead of Maven 3, which has a different way to use absolute paths within the project. 

This PR downpatches down to Maven 3.


TwoLettuce (approved):
cool beans

PR #663, opened 2026-07-16
update jacoco jars for java 23

mewilker:
## Return of the Java 23
@TwoLettuce noticed that with the new Java 23 update, there is error output coming from Jacoco.
image
image

This PR updates the Jacoco Jars to version 0.8.15. This version supports up to Java 26. My code no longer produces errors.

In case in the future they need to be updated again, I pulled the CLI jar from the maven repository. The agent jar in the auto grader was extracted from the jar downloaded from the maven repository. 

Will reviewers please run something from at least Phase 3 and make sure they also get no more errors? I'm pretty sure I've fixed, but VS code makes it difficult to copy the correct resources.


ZakkeryDaRebel (approved):
Works for me, so this change is sucessful!

Issue #565, opened 2025-03-25
Student did not receive early credit for commit days


frozenfrank:
Based on an initial inspection into the error, I think the error comes from line 230 ([view in the PR](https://github.com/softwareconstruction240/autograder/pull/514/files#diff-bf135d21e50011d73e5bf15e17e1b34057471e9e8a9f3fc0d987ff62b3af8d89R230)):

https://github.com/softwareconstruction240/autograder/blob/daaa599e7b75fd0e9886bf3d86d72bda7aa533b8/src/main/java/edu/byu/cs/autograder/score/LateDayCalculator.java#L227-L231

It's possible that the due date was not set to the end of the day. This would prevent the student from receiving credit for submitting at 8pm the day before since it is only 4 hours before the start of the due date.

This relates to behavior originally introduced in #514.

Maillman:
@frozenfrank I'm glancing at this issue now, but I'm not sure I understand this issue. Can you clarify what the issue is here?

Maillman:
Okay, so I don't see the issue here, and I am not aware of any issues that could arise from this behavior. So I'm going to go ahead and close this as not planned.

PR #667, opened 2026-08-13
Grace days calculator

mewilker:
## Overview
Resolves #664

Late Penalties have changed. Instead of a XX% up to a max of X days policy, the autograder now coordinates with Canvas to give/take "Grace Days to a student." If a student turns in an assignment early, they are awarded Grace Days. If a student turns in an assignment late, Grace Days are subtracted from the total. If the number of days late is greater than the number of Grace Days a student has, they receive a zero and keep their Grace Days. Holidays and Weekends are still factored into this equation. The autograder keeps track of these Grace Days per phase, and recalculates when resubmissions are made, using the current Canvas score. This should allow for Professors to add/remove Grace Days, should the occasion arise where they need to.

Something we should probably note about this implementation is that Grace Days are sent before the score for the assignment is sent. This is not a problem, but it should be noted that if a student receives a git block on their assignment, _their grace days will be updated, but their score for the assignment will not_. Furthermore, the Autograder no longer tries to preserve higher rubric items. If the score is overall better it will send it, but if writing unit tests introduces code quality issues and the score is not higher, the score will not be sent. Students must improve all aspects of the assignment. Another thing is that the autograder has no indication of whether a manual passoff has occurred or not. After some discussion with @TwoLettuce about this, we've decided that the easiest thing to do is require that students passing off submit their most recent change to the autograder. That way if they decide to improve code quality or unit tests after their manual pass-off, it doesn't double subtract Grace Days. If we tell a student to make changes based on their submission, they will need to resubmit to the autograder before attempting another manual pass-off.

## Details

- Autograder now keeps track of the Canvas assignment called `Grace days`
- Applying late penalties moved to a separate class (`PercentPenaltyCalculator`)
- Implementation of `GraceDayPenaltyCalculator` 
- Autograder no longer throws RuntimeExceptions if there are no future Holidays
- `ScorePair` moved from `Scorer` to `Rubric`

## Testing

- [x] Added/updated unit tests (credit goes mostly to @TwoLettuce for this)
- [ ] Tested edge cases

## Future Work

- A nice admin view to easily switch between penalties would make switching more convenient
- `The LateDayCalculator` relies on the configurable Max Late Days to determine how many days late an assignment is. The Grace Day Policy has no Late Day Max. Either you have the days or you don't. This means that the `PercentPenaltyCalculator` should probably factor in the max late day Policy, not the `LateDayCalculator`
- Git Penalties follow a completely different style entirely, due to a need for manual review and approval. I do think it would be good to have the various penalty calculators follow one style, and the `Scorer` calculator be the class decide if/when each penalty is applied. Perhaps at the same time, we could look into sending Grace Days at the same time.
- I refactored some tests from the `Scorer` to the `PercentPenaltyCalculator`, the main purpose of which was to determine that the scorer gave the best score on a per-rubric item basis. While I believe the move was correct, there may have been some aspects of the Scorer that are no longer being tested as thoroughly (my main concern is the `passed` function and the `transformRubric` functions). We may need to look into this.
- We have some decent logging for this at the moment, but as we gain more confidence in this system, we can probably remove some of the logging.

## Shoutouts

Thank you to @TwoLettuce and @camchamb for working on this while I was away :)


src/main/java/edu/byu/cs/autograder/score/penalties/GraceDayPenaltyCalculator.java
```diff
@@ -77,24 +75,14 @@ public Submission applyPenalty(Rubric rubric, int daysAfterDue, GradingContext g
                 "", gradingContext);
     }
 
-    private String makeCommentNotes(GradingContext gradingContext, int changeInGraceDays) {
```
camchamb: Did we no longer want a record of changes made to students' grace days in the comments of the grace days assignment?

mewilker: I noticed it was coming up in the comments of the assignment they'd submit on, I figured it was a lot of duplicate notifications.

mewilker:
After some thought I've decided to disable the test that is causing issues.

The problem with this test is that there seems to be some system discrepancy that is causing it to fail in the github environment, and I can't replicate it in mine. I would say the code and tests are overly complex. Its responsibility is to count commits, days committed and flag abnormalities in commits that frankly, I've never seen students attempt before, and the average student probably wouldn't know where to begin to try. For these responsibilities, the code has a lot of functions and classes that overall make it hard to follow what is going on. Refactoring the GitPenalties into a simpler calculator would probably help this issue.

The test that is failing is giving students credit when they haven't committed the full amount of days required, but their submission is early so the autograder is supposed to give their commits the green light anyways. There was an issue, #565 that documented this somewhat. However, they also couldn’t replicate the issue. I believe this is somewhat related.

The systems seem generally separate, so I'm genuinely not sure what changes made here broke this test, but ultimately, I think Grace Days is more important to have than passing off students early who didn't commit on enough days. We can simply pass them off manually.

